PR #26, opened 2022-12-01
Adding ability to supply a specific 'source' on the command line for content and package stores as well as telemetry to support proxy API scenario requirements.

brdeyo:
Current requirements for offsite manufacturing facility scenarios requires that the data (logs) and telemetry output by VC be uploaded to different data stores (by vendor). To enable support for this, the user/automation must be able to supply instructions to VC on the command line. We will be using the --parameters option to provide this information to the proxy interfacing classes (blob manager and logger).

### DIFF
--- a/src/VirtualClient/VirtualClient.Main/RunBootstrapCommand.cs
+++ b/src/VirtualClient/VirtualClient.Main/RunBootstrapCommand.cs
@@ -23,11 +23,6 @@ namespace VirtualClient
         public string ExperimentId { get; set; }
 
         /// <summary>
-        /// Metadata properties (key/value pairs) supplied to the application.
-        /// </summary>
-        public IDictionary<string, IConvertible> Metadata { get; set; }
-
-        /// <summary>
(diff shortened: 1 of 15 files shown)


src/VirtualClient/VirtualClient.Main/RunBootstrapCommand.cs
```diff
@@ -22,11 +22,6 @@ public class RunBootstrapCommand : CommandBase
         /// </summary>
         public string ExperimentId { get; set; }
 
-        /// <summary>
-        /// Metadata properties (key/value pairs) supplied to the application.
-        /// </summary>
-        public IDictionary<string, IConvertible> Metadata { get; set; }
-
```
yangpanMS, line 29: Do we not need metadata when bootstrapping anymore?

yangpanMS: Or they will still be supplied to telemetry

yangpanMS (approved):

PR #28, opened 2022-12-02
Update FIO Discovery and Multi-Throughput executors to use grouped job results.

brdeyo:
Historically, the --group_reporting option needs to be used with FIO to have the results of multiple jobs rollup into a single result. Otherwise, the results can produce disparate outcomes including measurements that are zeroes (0s).


src/VirtualClient/VirtualClient.Actions/DiskPerformanceWorkloadExecutor.cs
```diff
@@ -397,17 +397,16 @@ protected virtual IEnumerable<DiskPerformanceWorkloadProcess> CreateWorkloadProc
             return this.Logger.LogMessage($"{this.GetType().Name}.CreateProcesses", telemetryContext, () =>
             {
                 List<DiskPerformanceWorkloadProcess> processes = new List<DiskPerformanceWorkloadProcess>();
-                string testedInstance = this.DiskFilter;
 
                 if (string.Equals(processModel, WorkloadProcessModel.SingleProcess, StringComparison.OrdinalIgnoreCase))
                 {
-                    processes.Add(this.CreateWorkloadProcess(executable, commandArguments, testedInstance, disks.Select(disk => disk).ToArray()));
+                    processes.Add(this.CreateWorkloadProcess(executable, commandArguments, string.Join(",", disks.Select(d => d.DevicePath)), disks.ToArray()));
```
yangpanMS, line 403: The testedInstance was DiskFilter, if we do diskpath instead, we might need to sort it because the disk could come as different order, and then when we run the query it will come out different for different VM.

src/VirtualClient/VirtualClient.Contracts/VirtualClientComponentExtensions.cs
```diff
@@ -20,6 +19,64 @@ namespace VirtualClient.Contracts
     /// </summary>
     public static class VirtualClientComponentExtensions
     {
+        // Example Format:
+        // fio_{IOType}_{BlockSize}_{FileSize}_thmax{MaxThreads}
+        private static readonly Regex ParameterReferenceExpression = new Regex(@"(\x7B[\x20-\x7A\x7C\x7D-\x7E]+\x7D)|(\x5B[\x20-\x5A\x5C\x5E-\x7E]+\x5D)", RegexOptions.IgnoreCase | RegexOptions.Compiled);
```
yangpanMS, line 24: Could these be translated into literal characters? It will be more readable

yangpanMS (approved):

src/VirtualClient/VirtualClient.Contracts/VirtualClientComponentExtensions.cs
brdeyo: They can be but the Regex was a bit more flaky than I would like. This is why I went with the ASCII character references. The example above is meant to shed light on what an actual string looks like for the sake of readability.

yangpanMS (approved):

src/VirtualClient/VirtualClient.Actions/DiskPerformanceWorkloadExecutor.cs
brdeyo: I will change it back to the disk filter for now. My worry is that in doing so, we will not be able to distinguish variations of FIO in the future where there are different numbers of disks used. The results for running SingleProcess on 2 vs. 16 disks may not be equal but they will look equal in the data if we use the disk filter because the value will be the same.

yangpanMS (approved):

src/VirtualClient/VirtualClient.Actions/DiskPerformanceWorkloadExecutor.cs
brdeyo: Per our conversation, I am using the {ProcessModel},{DiskFilter},{DiskCount} for the metric categorization/tested instance (e.g. SingleProcess,BiggestSize,16 or SingleProcessPerDisk,BiggestSize,16).

src/VirtualClient/VirtualClient.Actions/DiskPerformanceWorkloadExecutor.cs
```diff
@@ -397,17 +397,20 @@ protected virtual IEnumerable<DiskPerformanceWorkloadProcess> CreateWorkloadProc
...
+                    processes.Add(this.CreateWorkloadProcess(executable, commandArguments, $"{WorkloadProcessModel.SingleProcess},{this.DiskFilter},{disks.Count()}", disks.ToArray()));
                 }
                 else if (string.Equals(processModel, WorkloadProcessModel.SingleProcessPerDisk, StringComparison.OrdinalIgnoreCase))
                 {
+                    // Example Metric Categorization
+                    // SingleProcessPerDisk,BiggestSize,16
                     processes.AddRange(new List<DiskPerformanceWorkloadProcess>(disks.Select(disk =>
                     {
-                        return this.CreateWorkloadProcess(executable, commandArguments, testedInstance, disk);
+                        return this.CreateWorkloadProcess(executable, commandArguments, $"{WorkloadProcessModel.SingleProcessPerDisk},{this.DiskFilter},{disks.Count()}", disk);
```
yangpanMS, line 413: Is this intentional that you use single disk but add diskcount?

yangpanMS (approved):

src/VirtualClient/VirtualClient.Actions/DiskPerformanceWorkloadExecutor.cs
brdeyo: It was but I guess it will not skew anything if we have it as SingleProcessPerDisk,DiskFilter,1

brdeyo: Changed it to use "1" instead of disk count.

yangpanMS: Actually I think your way is better, it tells you the total amount

yangpanMS (approved):

PR #29, opened 2022-12-03
Fix bug in DataTableExtensions.GetMetrics logic for handling DBNull.Value values in rows.

brdeyo:
This bug was found when parsing MLC results. Some of the row values in the data table end up with DBNull.Value values. The original implementation attempts to parse this as a double and exception is thrown. This work did not make it into the repo before moving to open source.

### DIFF
--- a/src/VirtualClient/VirtualClient.Contracts/Parser/DataTableExtensions.cs
+++ b/src/VirtualClient/VirtualClient.Contracts/Parser/DataTableExtensions.cs
@@ -1,22 +1,22 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Data;
-using System.Linq;
-using System.Text;
-using System.Text.RegularExpressions;
-using VirtualClient;
-using VirtualClient.Contracts;
-
-// Copyright (c) Microsoft Corporation.
+﻿// Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
 namespace VirtualClient.Contracts
 {
+    using System;
+    using System.Collections.Generic;
+    using System.Data;
+    using System.Linq;
+    using System.Text.RegularExpressions;
+    using VirtualClient;
+
     /// <summary>
     /// Extensions for parsing test documents related to data tables.
     /// </summary>
     public static class DataTableExtensions
     {
+        private static readonly Regex WhitespaceExpression = new Regex(@"\s+", RegexOptions.Compiled);
+
         /// <summary>
         /// Format text into data table delimited by column Delimiter regex.
         /// </summary>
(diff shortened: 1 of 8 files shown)


src/VirtualClient/VirtualClient.Contracts/Parser/DataTableExtensions.cs
```diff
@@ -138,7 +138,7 @@ public static DataTable ConvertToDataTable(string text, IList<KeyValuePair<int,
         /// <param name="tags">Tags</param>
         /// <param name="ignoreFormatError">If this method ignore FormatException. Default is false.</param>
         /// <returns>Parsed Metrics from a data table.</returns>
-        public static IList<Metric> GetMetric(
+        public static IList<Metric> GetMetrics(
```
yangpanMS, line 141: Just a reminder to us both, to update internal extension repo on this package

src/VirtualClient/VirtualClient.Contracts/Parser/DataTableExtensions.cs
```diff
@@ -161,37 +161,48 @@ public static DataTable ConvertToDataTable(string text, IList<KeyValuePair<int,
             {
                 try
                 {
-                    if (unitIndex >= 0)
+                    object value = row[valueIndex];
+                    if (value == null || value == DBNull.Value)
```
yangpanMS, line 165: Good catch, I didn't get an example of a null row. Would you think it's reasonable we report double.NaN?

yangpanMS: This will tell us we get a null cell, vs ignored, we will not know if we got a row or not

yangpanMS (approved):

src/VirtualClient/VirtualClient.Contracts/Parser/DataTableExtensions.cs
brdeyo: I'd rather just ignore. It means the data was not good and we have the raw results to consult if we need.

src/VirtualClient/VirtualClient.Contracts/Parser/DataTableExtensions.cs
brdeyo: Yep. I will be following up this PR merger with the updates to our ADO repo.

src/VirtualClient/VirtualClient.Contracts/Parser/DataTableExtensions.cs
yangpanMS: The SAP workloads has tons of dashes that we are reporting double.NaN. They indicated they much prefer double.NaN vs ignore. Because they can see that machine was there doing nothing.

PR #33, opened 2022-12-20
Enable the ability for users to define custom ports for the VC REST API service. Documentation updates to modify ordering.

brdeyo:
Went back and forth on making the API ports something defined on the command line or in the profiles. I tried out both of these avenues and generally preferred defining them on the command line leaving ports associated with workloads (e.g. NTttcp, CPS) to be defined in the workload profiles.


website/docs/developing/ci-cd.md
```diff
@@ -1,9 +1,8 @@
 ﻿---
-id: ci-cd
-sidebar_position: 8
+sidebar_position: 45
 ---
 
-# Build and CI / CD pipeline
+# Build a CI / CD pipeline
```
yangpanMS, line 5: It was meant as dotnet Build and CI/CD pipeline and not "build a pipeline" but don't worry about it

yangpanMS: ok let me know if you want me to show you how to debug locally

PR #42, opened 2023-01-10
Implement Downloading a Blob By Chunks via Range Processing

cjhillbrand:
This PR introduces functionality for the Virtual Client to download a Blob from a Proxy API by downloading the blob by chunks. 
If the downstream Proxy API does not support range processing, then Virtual Client will attempt to download the file as originally implemented, downloading the entire blob in one go.
The download follows this sequence:
1. Perform HEAD on the Blob Resource.
2. Check if the headers required for range processing are present as well as the HEAD response returns a success status code.
3. With a pre-defined chunk size (1MiB) the Virtual Client issues a series of requests to the Proxy API and iteratively builds the blob to a provided stream.


src/VirtualClient/VirtualClient.Core/Proxy/ProxyBlobManager.cs
```diff
@@ -88,11 +88,9 @@ public async Task<DependencyDescriptor> DownloadBlobAsync(DependencyDescriptor d
                 blobDescriptor.ContentEncoding.WebName,
                 blobPath);
 
-            HttpResponseMessage response = await this.ApiClient.DownloadBlobAsync(info, downloadStream, cancellationToken)
+            await this.ApiClient.DownloadBlobAsync(info, downloadStream, cancellationToken)
                 .ConfigureAwait(true);
 
-            response.ThrowOnError<DependencyException>(ErrorReason.DependencyInstallationFailed);
```
yangpanMS, line 94: Is there a reason we are taking the error reporting out? Did you move this responsibility in the apiclient? It's not clear to me that you are letting client reporting all errors

src/VirtualClient/VirtualClient.Core.UnitTests/Proxy/ProxyBlobManagerTests.cs
```diff
@@ -187,36 +182,9 @@ public Task ProxyBlobManagerDownloadsTheExpectedBlobThroughTheProxyApi_Scenario2
...
                     Assert.AreEqual(Encoding.ASCII.WebName, blobDescriptor.ContentEncoding);
-                })
-                .ReturnsAsync(this.mockFixture.CreateHttpResponse(System.Net.HttpStatusCode.OK));
-
-                return blobManager.DownloadBlobAsync(descriptor, this.mockStream, CancellationToken.None);
-            }
-        }
-
-        [Test]
-        public void ProxyBlobManagerThrowsIfTheBlobDownloadFails()
```
yangpanMS, line 198: Is there a reason you are changing this behavior? Please check Bryan's latest PR on proxy blob manager changes

cjhillbrand: added back in

src/VirtualClient/VirtualClient.Core/Proxy/ProxyBlobManager.cs
cjhillbrand: added back in

yangpanMS (approved):

PR #45, opened 2023-01-12
expanding openSSL coverage

ericavella:
primarily to add rsa algorithms
also adding aes cipher with different & popular encoding, sha3 digest, aria, blowfish, idea ciphers


ericavella:
btw -- i know it looks like the profile was completely erased and rewritten 😨 i was having an issue with indentation in VS while i was adding new objects to the json

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-OPENSSL.json
```diff
@@ -1,117 +1,225 @@
...
+      "Type": "OpenSslExecutor",
+      "Parameters": {
+        "Scenario": "AES-256-CFB",
+        "CommandArguments": "speed -elapsed -seconds 100 aes-256-cfb",
+        "PackageName": "openssl",
+        "Tags": "CPU,OpenSSL,Cryptography"
+      }
+    },
+    {
+      "Type": "OpenSslExecutor",
```
yangpanMS, line 123: If I remember correctly CBC algorithms have known security vulnerabilities, could you do a little research on that?

yangpanMS:
Could you get an example output with the new profiles, and make sure our parser handles all the new algorithms? Thanks!

yangpanMS (approved):
Great job! You might already have done it but please build a new package, run on a VM and make sure all the metrics from past and new are there.

ericavella:
tested on VM and confirmed all past & present metrics exist

PR #47, opened 2023-01-20
Adding Provision for Kill and SafeKill entireProcessTree

imadityaa:
The objective is to kill all associated child/descendent processes that have been created by VC/workload, when VC timeouts or encounters a cancellation token


src/VirtualClient/VirtualClient.TestFramework/InMemoryProcess.cs
```diff
@@ -146,6 +146,14 @@ public void Kill()
             this.OnKill?.Invoke();
         }
 
+        /// <summary>
+        /// Kills the fake process, and associated child processes.
+        /// </summary>
+        public void Kill(bool entireProcessTree)
```
yangpanMS, line 152: I would make it into the original method and make it a default. When VC kills threads, we almost always want to kill the entire tree

yangpanMS (commented):
This looks good, but let's change the original method and make it like this
public void Kill(bool entireProcessTree = true)

src/VirtualClient/VirtualClient.TestFramework/InMemoryProcess.cs
imadityaa: Made it "true" by default in process.Kill(), in ProcessProxy.cs. Thus, process.Kill() is same as process.kill(true). It will be applied to all workloads. In case somewhere false is needed explicitly, the process.Kill(false) or process.SafeKill(false) can be used.

brdeyo (approved):

yangpanMS (approved):

PR #48, opened 2023-01-20
Fix issue where VC deletes mount path when testing raw disk

yangpanMS:


src/VirtualClient/VirtualClient.Actions/FIO/FioDiscoveryExecutor.cs
```diff
@@ -324,6 +324,8 @@ await this.CleanUpWorkloadTestFilesAsync()
         protected override DiskPerformanceWorkloadProcess CreateWorkloadProcess(string executable, string commandArguments, string testedInstance, params Disk[] disksToTest)
         {
             string[] testFiles = disksToTest.Select(disk => disk.DevicePath).ToArray();
+            // Since in this case we are testing on raw disks, we are not cleaning up test files
+            this.DeleteTestFilesOnFinish = false;
```
brdeyo, line 328: Put this definition in the constructor for Discovery and for Multi-Throughput as well I believe. I think they both target raw disk.

brdeyo (approved):

src/VirtualClient/VirtualClient.Actions/FIO/FioDiscoveryExecutor.cs
yangpanMS: done

PR #54, opened 2023-01-31
Addressing a few design flaws in the MockFixture and InMemoryApiClient implementations causing object disposed exceptions.

brdeyo:
The HttpResponseMessage objects being created by the mock classes when reused can hit object disposed exceptions. This is not expected given the default mock setups.


brdeyo:
These changes are minor and we need to get a new version of VC out today for production release qualification runs.

PR #61, opened 2023-02-21
sysbench client IP address bug

ericavella:
bug fix, gets the correct client IP address now
built a state check into the server executor as well

fixes the bug from this error during QoS VC check: https://msazure.visualstudio.com/One/_backlogs/backlog/CRC-VirtualClient/Features/?workitem=17132374


src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPServerExecutor.cs
```diff
@@ -46,7 +46,17 @@ protected override async Task InitializeAsync(EventContext telemetryContext, Can
         {
             await base.InitializeAsync(telemetryContext, cancellationToken).ConfigureAwait(false);
             this.InitializeApiClients();
-            this.ConfigureMySQLPrivileges(cancellationToken);
+
+            SysbenchOLTPState state = await this.StateManager.GetStateAsync<SysbenchOLTPState>($"{nameof(SysbenchOLTPState)}", cancellationToken)
+                ?? new SysbenchOLTPState();
+
+            if (!state.MySQLDBInitialized)
```
yangpanMS, line 53: Could you add unit test to verify the executor would do the right thing for different state? You could mock the state object and verify the SQL will be configured if not initialized, and will be skipped if initialized.

src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPServerExecutor.cs
```diff
@@ -100,15 +110,18 @@ await process.WaitForExitAsync(cancellationToken)
             });
         }
 
-        private async Task ConfigureMySQLPrivileges(CancellationToken cancellationToken)
+        private async Task ConfigureMySQLPrivilegesAsync(CancellationToken cancellationToken)
         {
             string workingDirectory = this.GetPackagePath(this.PackageName);
 
+            string dropUserCommand = $"mysql --execute=\"DROP USER 'sbtest'@'{this.ClientIpAddress}'\"";
```
yangpanMS, line 117: Where did the "sbtest" user name come from? If it's from the dependency class, make a parameter on both dependency and action

ericavella: oooooh! okay i wasn't sure if i was able to retrieve client/server ip addresses from the mysql dependency directly. let me reconfigure all this then

yangpanMS (commented):
Could you check with Bryan and see if your changes on the DB setup will conflict with his? I know he is doing some changes to this workload on DB setup.

ericavella:
sorry for the messy PR! this fixes the bug with minimal changes

yangpanMS (approved):

PR #65, opened 2023-03-07
Users/nmalkapuram/integrating linpack

nmalkapuram:


brdeyo (changes requested):

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPL.json
```diff
@@ -0,0 +1,45 @@
+{
+  "Description": "High Performance Linkpack Performance workload",
+  "Parameters": {
+    "AccountKey": "[secret:commandline]=StorageAccountKey"
```
yangpanMS, line 4: This is not needed anymore, if you copied it from some other profile, you could delete from there too

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPL.json
```diff
@@ -0,0 +1,45 @@
...
+        "CompilerName": "gcc",
+        "CompilerVersion": "11"
+      }
+    },
+    {
+      "Type": "DependencyPackageInstallation",
+      "Parameters": {
+        "Scenario": "InstallARMPerformanceLibrariesPackage",
+        "BlobContainer": "packages",
+        "BlobName": "armperformancelibraries.zip",
```
yangpanMS, line 39: Could you document why we need this for ARM only? And what's in this package?

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPL.json
```diff
@@ -0,0 +1,45 @@
...
+        "PackageName": "arm_performance_libraries",
```
yangpanMS, line 40: If this is referenced in your executor, please make that also a parameter in your executor

src/VirtualClient/VirtualClient.Actions/HPL/HPLMetricsParser.cs
```diff
@@ -0,0 +1,135 @@
...
+                throw new SchemaException("The HPLinpack output file has incorrect format for parsing");
+            }
+            else
+            {
+                this.Metrics = new List<Metric>();
+                for (int i = 0; i < matches.Count; i++)
+                {
+                    var st = Regex.Split(matches.ElementAt(i).Value, SplitAtSpace);
+
+                    /*this.TotalThroughput += Convert.ToDouble(st[1]);
```
yangpanMS, line 73: Please clean up these comments once finished

yangpanMS (changes requested):
I see this is still work in progress, please ping me when you are done. Please also add the documents, and the licenses link.

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPL.json
nmalkapuram: Any specific reason why we are moving away from having account key parameters in profile?

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPL.json
nmalkapuram: Yep done. Adding it in documentation. It has core math libraries that are specific to ARM processors so that is part of dependencies to run HPL.

src/VirtualClient/VirtualClient.Actions.UnitTests/HPLinpack/HPLinpackExecutorTests.cs
```diff
@@ -0,0 +1,212 @@
...
+
+            using (TestHPLExecutor executor = new TestHPLExecutor(this.fixture))
+            {
+                int executedArmCommands = 0;
+
+                this.fixture.ProcessManager.OnCreateProcess = (command, arguments, workingDirectory) =>
+                {
+                    if (expectedCommandsOnLinuxarm64.Any(c => c == $"{command} {arguments}"))
+                    {
+                        executedArmCommands++;
```
brdeyo, line 86: This is not a definitive way to check for the expected commands to have been executed. A more precise technique is to simply remove the commands from the set of expectedCommands and then Assert.IsEmpty(expectedCommands) at the end.

In fact, the most precise way that validates the expected commands were executed and in the exact order expected, you remove from expectedCommands[0] ONLY each time in the OnCreateProcess.

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,452 @@
...
+            this.systemManagement = this.Dependencies.GetService<ISystemManagement>();
+            this.packageManager = this.systemManagement.PackageManager;
+            this.fileSystem = this.systemManagement.FileSystem;
+            this.stateManager = this.systemManagement.StateManager;
+        }
+
+        /// <summary>
+        /// The order of the coefficient matrix also known as problem size (N)
+        /// </summary>
+        public string N
```
brdeyo, line 54: Let's spell this out a bit more clearly. I know that Linpack may use 'N' to describe this but it would be better with a name such as ProblemSize or ProblemSizeN

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,452 @@
...
+            set
+            {
+                this.Parameters[nameof(this.N)] = value;
+            }
+        }
+
+        /// <summary>
+        /// The partitioning blocking factor also known as block size (NB)
+        /// </summary>
+        public string NB
```
brdeyo, line 71: Same comment as above. Call this the PartitionBlockingFactor or PartitionBlockingFactorNB

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,452 @@
...
+            {
+                this.Parameters.TryGetValue(nameof(HPLinpackExecutor.HPLVersion), out IConvertible hplversion);
+                return hplversion?.ToString();
+            }
+        }
+
+        /// <summary>
+        /// The path where the HPL directory.
+        /// </summary>
+        public string HPLDirectory { get; set; }
```
brdeyo, line 112: Properties that are not parameter readers should be protected or private member variables of the class.

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,452 @@
...
+                .ConfigureAwait(false);
+            }
+
+            DateTime endTime = DateTime.UtcNow;
+            this.LogMetrics(results, startTime, endTime, telemetryContext, cancellationToken);
+        }
+
+        private void SetParameters()
+        {
+            if (this.N == null)
```
brdeyo, line 199: This can be set in the property at the top passing in a default value.

this.Parameters.GetValue<int>(nameof(.....), Environment.ProcessorCount * 10000)

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,452 @@
...
+
+        }
+
+        private void ThrowIfPlatformIsNotSupported()
+        {
+            if (this.Platform != PlatformID.Unix)
+            {
+                throw new WorkloadException(
+                    $"The HPL workload is currently only supported on the following platform/architectures: " +
+                    $"'{PlatformSpecifics.GetPlatformArchitectureName(PlatformID.Unix, Architecture.X64)}', '{PlatformSpecifics.GetPlatformArchitectureName(PlatformID.Unix, Architecture.Arm64)}'. ",
```
brdeyo, line 228: Do me a favor and let's add the following public static properties to the PlatformSpecifics class. I keep seeing a need for the platform architecture names in error messages. This will allow us to shorten the length of some of these error messages and we can use string.Join() techniques for ease-of-coding as well.

class PlatformSpecifics
{
      public static readonly LinuxX64 = PlatformSpecifics.GetPlatformArchitectureName(PlatformID.Unix, ....);
      public static readonly LinuxArm64 = PlatformSpecifics.GetPlatformArchitectureName(PlatformID.Unix, ....);
      public static readonly WinX64 = PlatformSpecifics.GetPlatformArchitectureName(PlatformID.Win32NT, ....);
      public static readonly WinArm64 = PlatformSpecifics.GetPlatformArchitectureName(PlatformID.Win32NT, ....);
}

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,452 @@
...
+
+                await this.fileSystem.File.ReplaceInFileAsync(
+                        makeFilePath, @"LAinc *=", "LAinc = -I/usr/lib/x86_64-linux-gnu", cancellationToken);
+            }
+
+            await this.fileSystem.File.ReplaceInFileAsync(
+                    makeFilePath, @"CC *= *[^\n]*", "CC = mpicc", cancellationToken); 
+        }
+
+        private async Task ConfigureHPLdatFileAsync(EventContext telemetryContext, CancellationToken cancellationToken)
```
brdeyo, line 277: Naming: ConfigureDatFileAsync

Unless there is another .dat file involved in the operations workflow, there is not a real need to call out the HPL part given the developer is working in an executor that makes that part relatively clear.

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,452 @@
...
+                        output = process.StandardOutput.ToString();
+
+                    }
+                }).ConfigureAwait(false);
+            }
+
+            return output;
+        }
+
+        private async Task<string> ExecuteElevatedCommandAsync(string pathToExe, string commandLineArguments, string workingDirectory, EventContext telemetryContext, CancellationToken cancellationToken)
```
brdeyo, line 372: Just pass in an optional boolean to the ExecuteCommandAsync method above. No need for 2 methods.

private async Task<string> ExecuteCommandAsync(string pathToExe, string commandLineArguments, string workingDirectory, EventContext telemetryContext, CancellationToken cancellationToken, bool runElevated = false)

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
```diff
@@ -0,0 +1,48 @@
...
+  "Metadata": {
+    "RecommendedMinimumExecutionTime": "00:01:00",
+    "SupportedPlatforms": "linux-x64,linux-arm64",
+    "SupportedOperatingSystems": "Ubuntu"
+  },
+  "Actions": [
+    {
+      "Type": "HPLinpackExecutor",
+      "Parameters": {
+        "Scenario": "ProcessorSpeed",
```
yangpanMS, line 12: In the scenario, could you include the block size and  problem size? Or do you have another way to distinguish the data? I don't see it in the parser

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,408 @@
...
+        protected override async Task InitializeAsync(EventContext telemetryContext, CancellationToken cancellationToken)
+        {
+            this.ThrowIfPlatformIsNotSupported();
+
+            if (this.CpuArchitecture == Architecture.Arm64)
+            {
+                DependencyPath armPerformanceLibrariesPackage = await this.packageManager.GetPackageAsync(this.ARMPerformanceLibrariesPackageName, cancellationToken)
+                .ConfigureAwait(false);
+                string armPackageLibrariesPath = armPerformanceLibrariesPackage.Path;
+                await this.systemManagement.MakeFileExecutableAsync(this.PlatformSpecifics.Combine(armPackageLibrariesPath, "arm-performance-libraries_22.1_Ubuntu-20.04.sh"), this.Platform, cancellationToken).ConfigureAwait(false);
```
yangpanMS, line 125: Is this specific to ARM? Could you do a check for the OS, and also include link/doc on why we need to do this

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
```diff
@@ -0,0 +1,48 @@
...
+        "Scenario": "InstallRequiredLinuxPackagesForHPL",
+        "Repositories-Apt": "ppa:ubuntu-toolchain-r/test",
+        "Packages": "build-essential,linux-generic,libmpich-dev,libopenmpi-dev,libatlas-base-dev"
+      }
+    },
+    {
+      "Type": "CompilerInstallation",
+      "Parameters": {
+        "CompilerName": "gcc",
+        "CompilerVersion": "11"
```
yangpanMS, line 34: Let's make the compiler version an overwritable parameter.

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,408 @@
...
+
+            string results;
+            if (this.CpuArchitecture == Architecture.X64)
+            {
+                results = await this.ExecuteCommandAsync("runuser", $"-u azureuser -- mpirun --use-hwthread-cpus -np {Environment.ProcessorCount} ./xhpl", this.PlatformSpecifics.Combine(this.HPLDirectory, "bin", "Linux_GCC"), true, telemetryContext, cancellationToken)
+                .ConfigureAwait(false);
+            }
+            else
+            {
+                results = await this.ExecuteCommandAsync("runuser", $"-u azureuser -- mpirun -np {Environment.ProcessorCount} ./xhpl", this.PlatformSpecifics.Combine(this.HPLDirectory, "bin", "Linux_GCC"), true, telemetryContext, cancellationToken)
```
yangpanMS, line 178: Parameterize the username, apply throughout

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,408 @@
...
+
+            DateTime endTime = DateTime.UtcNow;
+            this.LogMetrics(results, startTime, endTime, telemetryContext, cancellationToken);
+        }
+
+        private void SetParameters()
+        {
+            // gives you P*Q = Environment.ProcessorCount and P <= Q and  Q-P to be the minimum possible value.
+            this.ProcessRows = 1;
+            this.ProcessColumns = Environment.ProcessorCount;
```
yangpanMS, line 190: Get this processor count from systemmanager so that you could mock and test this

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,408 @@
...
+            for (int i = 2; i <= Math.Sqrt(Environment.ProcessorCount); i++)
```
yangpanMS, line 191: Same here, get from system manager, apply throughout

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,408 @@
...
+                throw new WorkloadException(
+                    $"The HPL workload is currently only supported on the following platform/architectures: " +
+                    $"'{PlatformSpecifics.LinuxX64}', '{PlatformSpecifics.LinuxArm64}'.",
+                    ErrorReason.PlatformNotSupported);
+            }
+        }
+
+        private async Task ConfigureMakeFileAsync(EventContext telemetryContext, CancellationToken cancellationToken)
+        {
+            string makeFilePath = this.PlatformSpecifics.Combine(this.HPLDirectory, "setup", "Make.Linux_GCC");
```
yangpanMS, line 218: parameterize the Make.Linux_GCC into something like string configName/makefileName. This happens a lot

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,408 @@
...
+                        makeFilePath, @"LAinc *=", $"LAinc = $(ARMPL_INCLUDES)", cancellationToken);
+
+                await this.fileSystem.File.ReplaceInFileAsync(
+                        makeFilePath, @"LAlib *= *[^\n]*", "LAlib = /opt/arm/armpl_22.1_gcc-11.2/lib/libarmpl.a", cancellationToken);
+
+                await this.fileSystem.File.ReplaceInFileAsync(
+                        makeFilePath, @"LINKER *= *[^\n]*", "LINKER = mpifort", cancellationToken);
+
+                await this.fileSystem.File.ReplaceInFileAsync(
+                        makeFilePath, @"CCFLAGS *= *[^\n]*", "CCFLAGS = $(HPL_DEFS) -Ofast -march=armv8-a", cancellationToken);
```
yangpanMS, line 242: Let's parameterize the cflags, cflags might not be the standard.

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,408 @@
...
+                        makeFilePath, @"MPinc *=", $"MPinc =  -I/usr/lib/x86_64-linux-gnu/openmpi", cancellationToken);
+
+                await this.fileSystem.File.ReplaceInFileAsync(
+                        makeFilePath, @"MPlib *=", "MPlib =  /usr/lib/x86_64-linux-gnu/openmpi/lib/libmpi.so", cancellationToken);
+
+                await this.fileSystem.File.ReplaceInFileAsync(
+                        makeFilePath, @"LAinc *=", "LAinc = -I/usr/lib/x86_64-linux-gnu", cancellationToken);
+            }
+
+            await this.fileSystem.File.ReplaceInFileAsync(
```
yangpanMS, line 256: Is there a cflag replacement with the linux-x64? I don't see it

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackMetricsParser.cs
```diff
@@ -0,0 +1,78 @@
...
+
+                    Dictionary<string, IConvertible> metadata = new Dictionary<string, IConvertible>()
+                         {
+                             { "N_WR{st[0]}", st[1] },
+                             { "NB_WR{st[0]}", st[2] },
+                             { "P_WR{st[0]}", st[3] },
+                             { "Q_WR{st[0]}", st[4] },
+                         };
+
+                    this.Metrics.Add(new Metric($"Time", Convert.ToDouble(st[5]), "secs", metadata: metadata));
```
yangpanMS, line 70: Are these N_WR metrics useful? If so, I don't recommend putting them into metadata, they should be proper metrics. If they are not as useful, please at least check them in unit tests.

website/docs/workloads/HPLinpack/HPLinpack-profiles.md
```diff
@@ -0,0 +1,40 @@
...
+* [Workload Profile](https://github.com/microsoft/VirtualClient/blob/main/src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json) 
+
+* **Supported Platform/Architectures**
+  * linux-x64
+  * linux-arm64
+
+* **Supports Disconnected Scenarios**  
+  * No. Internet connection required.
+
+* **Dependencies**
```
yangpanMS, line 22: The dependency part is for actually explaining the dependencies we put in the profile, like compiler installation, and the ARM package. Please put link, and how/why we need that ARM binaries to run Linpack

yangpanMS (changes requested):

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackMetricsParser.cs
nmalkapuram: N, NB, P,Q are the input metrics. We can add it as part of scenario, but having them as separate metrics doesn't add much value.

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
nmalkapuram: Nope we don't default CCFLAG value generated in HPL.dat file while hpl build works well in linux-x64 case.

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
nmalkapuram: I have added it in the metadata for each metric generated. I could also add them in scenario to be more clear.

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackMetricsParser.cs
yangpanMS: ok makes sense then, just verify them in the unit tests, one of each would be enough validation

src/VirtualClient/VirtualClient.Actions.UnitTests/HPLinpack/HPLinpackExecutorTests.cs
```diff
@@ -0,0 +1,209 @@
...
+            {
+                List<string> expectedCommands = new List<string>()
+                {
+                    $"wget http://www.netlib.org/benchmark/hpl/hpl-{this.fixture.Parameters["Version"]}.tar.gz -O {this.fixture.Parameters["PackageName"]}.tar.gz",
+                    $"tar -zxvf {this.fixture.Parameters["PackageName"]}.tar.gz",
+                    $"sudo bash -c \"source make_generic\"",
+                    $"mv Make.UNKNOWN Make.Linux_GCC",
+                    $"ln -s {this.fixture.PlatformSpecifics.Combine(executor.GetHPLDirectory, "setup", "Make.Linux_GCC" )} Make.Linux_GCC",
+                    $"make arch=Linux_GCC",
+                    $"sudo runuser -u {this.fixture.Parameters["Username"]} -- mpirun --use-hwthread-cpus -np 16 ./xhpl"
```
yangpanMS, line 119: why would there be a space between -- and mpirun?

src/VirtualClient/VirtualClient.Actions.UnitTests/HPLinpack/HPLinpackMetricsParserTests.cs
```diff
@@ -0,0 +1,63 @@
...
+    using NUnit.Framework;
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using System.Reflection;
+    using System.Text;
+    using System.Threading.Tasks;
+    using VirtualClient.Contracts;
+    using Microsoft.Identity.Client;
```
yangpanMS, line 16: Is the SchemaException coming from this namespace? Is there a normal schema exception you could use?

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,434 @@
...
+        }
+
+        /// <summary>
+        /// The order of the coefficient matrix also known as problem size (N)
+        /// </summary>
+        public string ProblemSizeN
+        {
+            get
+            {
+                return this.Parameters.GetValue<string>(nameof(HPLinpackExecutor.ProblemSizeN), Environment.ProcessorCount * 10000);
```
yangpanMS, line 72: Again, please get this from systemmanager

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -0,0 +1,434 @@
...
+            await this.ExecuteCommandAsync("make", $"arch=Linux_GCC", this.HPLDirectory, false, telemetryContext, cancellationToken)
+                .ConfigureAwait(false);
+
+            this.SetParameters();
+            await this.ConfigureDatFileAsync(telemetryContext, cancellationToken).ConfigureAwait(false);
+
+            string results;
+            if (this.CpuArchitecture == Architecture.X64)
+            {
+                results = await this.ExecuteCommandAsync("runuser", $"-u {this.Username} -- mpirun --use-hwthread-cpus -np {this.coreCount} ./xhpl", this.PlatformSpecifics.Combine(this.HPLDirectory, "bin", "Linux_GCC"), true, telemetryContext, cancellationToken)
```
yangpanMS, line 199: Are the spaces between -- and mpirun intentional?

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
nmalkapuram: My bad. Fixing it right back.

src/VirtualClient/VirtualClient.Actions.UnitTests/HPLinpack/HPLinpackExecutorTests.cs
nmalkapuram: -- is treated as parameter just like -c or --command and "mpirun --use-hwthread-cpus -np 16 ./xhpl" is the value to that parameter.

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
nmalkapuram: runuser [options] -u <user> [[--] <command>]
runuser [options] [-] [<user> [<argument>...]]

Run <command> with the effective user ID and group ID of <user>. 
The options -c, -f, -l, and -s are mutually exclusive with -u.
where - c is command , -f,-l,-s are few other parameters for runuser. It's the syntax of runuser and we can't use -u and -c parameters at once. These are mutually exclusive.

src/VirtualClient/VirtualClient.Actions.UnitTests/HPLinpack/HPLinpackMetricsParserTests.cs
nmalkapuram: It's from VirtualClient namespace.

brdeyo (approved):

website/docs/workloads/HPLinpack/HPLinpack-profiles.md
```diff
@@ -0,0 +1,66 @@
+﻿# HPLinpack Workload Profiles
```
brdeyo, line 1: Let's stick with the casing conventions we have in place for the docs if you don't mind. It matters for ensuring we don't run into casing issues in the URIs. As such, the folder names should be lower-cased as should the file names. You can look at the other folders in the /docs/workloads directory for examples.  Thanks!

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
```diff
@@ -0,0 +1,59 @@
...
+        "CompilerName": "$.Parameters.CompilerName",
+        "CompilerVersion": "$.Parameters.CompilerVersion"
+      }
+    },
+    {
+      "Type": "DependencyPackageInstallation",
+      "Parameters": {
+        "Scenario": "InstallHPLinpackPackage",
+        "BlobContainer": "packages",
+        "BlobName": "hpl.2.3.zip",
```
yangpanMS, line 53: I just need a confirmation that in this package, you included the original license from linpack

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
```diff
@@ -0,0 +1,59 @@
...
+        "PackageName": "hplinpack",
```
yangpanMS, line 54: Did you use a .vcpkg file when zipping the package? Is that checked in this branch?

website/docs/workloads/hplinpack/hplinpack-profiles.md
```diff
@@ -0,0 +1,68 @@
+﻿# HPLinpack Workload Profiles
+The following profiles run customer-representative or benchmarking scenarios using the HPLinpack workload.  
+
+* [Workload Details](./HPLinpack.md)
```
yangpanMS, line 4: nit:lower case please

yangpanMS (approved):
If you used a vcpkg file, please add it to the source

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
nmalkapuram: Yes. It has copyright file which has copyright notice and license terms information.

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
nmalkapuram: No haven't used that .vcpkg file while zipping.

yangpanMS (approved):

PR #71, opened 2023-03-28
mysql server error fix

ericavella:
couldn't initialize mysql privileges w/o packages directory
also after that needed 'if exists' for the drop user command or else would throw error on first iteration

### DIFF
--- a/src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPServerExecutor.cs
+++ b/src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPServerExecutor.cs
@@ -103,8 +103,9 @@ namespace VirtualClient.Actions
         private async Task ConfigureMySQLPrivilegesAsync(CancellationToken cancellationToken)
         {
             string workingDirectory = this.GetPackagePath(this.PackageName);
+            this.SystemManager.FileSystem.Directory.CreateDirectory(workingDirectory);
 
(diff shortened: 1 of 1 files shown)


yangpanMS (approved):
Approved, but do you need to update any of the tests for this? If not we should have a unit test guarding this

src/VirtualClient/VirtualClient.Actions.UnitTests/SysbenchOLTP/SysbenchOLTPServerExecutorTests.cs
```diff
@@ -0,0 +1,94 @@
...
+                if (expectedCommands.Any(c => c == $"{exe} {arguments}"))
+                {
+                    commandExecuted++;
+                }
+
+                return this.mockFixture.Process;
+            };
+
+            CancellationTokenSource cancellationTokenSource = new CancellationTokenSource();
+            cancellationTokenSource.CancelAfter(1000);
```
yangpanMS, line 75: If you can, limit that to couple milliseconds, we are just doing mock commands, so I imagine these can finish in 10 milliseconds

yangpanMS: Reason for this, is we have hundreds of unit tests, and if each test take 1 second, they add up, but couple milliseconds are ok

yangpanMS (approved):

PR #72, opened 2023-04-06
Replace spack installation with git clone

yangpanMS:


yangpanMS:
There is no reason that spack installation to be a class, git clone should be enough for this purpose

PR #76, opened 2023-04-10
sysbench resiliency in creating tables

ericavella:


src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPClientExecutor.cs
```diff
@@ -258,20 +255,27 @@ private Task ExecuteWorkloadAsync(IPAddress serverIpAddress, EventContext teleme
             {
                 using (BackgroundOperations profiling = BackgroundOperations.BeginProfiling(this, cancellationToken))
                 {
-                    this.sysbenchExecutionArguments = 
-                    $"{this.Workload} --threads={this.Threads} --time={this.DurationSecs} --tables={this.NumTables} --table-size={this.RecordCount} --mysql-db={this.DatabaseName} " +
-                    $"--mysql-host={this.ServerIpAddress}";
+                    string sysbenchPath = this.PlatformSpecifics.Combine(this.sysbenchDirectory, SysbenchOLTPClientExecutor.SysbenchFileName);
+                    this.sysbenchExecutionArguments += $"--mysql-host={this.ServerIpAddress} ";
+
+                    // For resiliency, cleanup the sysbench environment (ie. delete tables) in the case that a run failed before, and skipped cleanup
```
yangpanMS, line 261: Instead of cleaning up, can we do this:
Check if the table sbtest1 exists, if it exists, don't create that table

yangpanMS (approved):
I will approve this to unblock testing, but please research how to minimize table creation and recreation when we have larger data tables, thanks

PR #78, opened 2023-04-12
Updating username parameter in profile and version in azure-pipelines.yml

nmalkapuram:


src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
```diff
@@ -6,7 +6,7 @@
     "SupportedOperatingSystems": "Ubuntu"
   },
   "Parameters": {
-    "Username": "",
+    "Username": "testuser",
```
yangpanMS, line 9: Hey @nmalkapuram I made the change that the username will default to the Environment.UserName if the profile doesn't specify one, is that working? Do you have to make this change?

nmalkapuram: It wasn't working. It's taking as null value, if isn't supplied from command line.

yangpanMS (approved):
I will unblock your experiment for now. Ideally we should leave profile blank and let Enviroment.CurrentUser do the work

PR #83, opened 2023-04-17
Updating pip and setuptools

nmalkapuram:
Upgrading pip and setuptools.


yangpanMS (approved):
Thank you! Does this break any unit test? If not maybe we should add some. Also in the future let's also have a pip package manager

yangpanMS (approved):

PR #86, opened 2023-04-20
Adding SOC Stress

deep1712:


src/VirtualClient/VirtualClient.Main/profiles/PERF-SOC-STRESS.json
```diff
@@ -0,0 +1,42 @@
+{
+    "Description": "SOC Stress Workload",
+    "Parameters": {
+        "Host":"HostIP",
+        "UserName":"username",
```
brdeyo, line 5: Rename:
SocUserName
SocPassword

src/VirtualClient/VirtualClient.Main/profiles/PERF-SOC-STRESS.json
```diff
@@ -0,0 +1,42 @@
+{
+    "Description": "SOC Stress Workload",
+    "Parameters": {
+        "Host":"HostIP",
```
brdeyo, line 4: SocIPAddress

src/VirtualClient/VirtualClient.Common/ProcessDetails.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// Exit code of the command executed.
+        /// </summary>
+        public int ExitCode { get; set; }
+
+        /// <summary>
+        /// Generated Results of the command.
+        /// </summary>
+        public string GeneratedResults { get; set; }
```
brdeyo, line 26: Need to support more than 1 set of results. Some of the workloads produce multiple files worth of output for a single run of the process.

src/VirtualClient/VirtualClient.Common/ProcessDetails.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// Standard error of the command.
+        /// </summary>
+        public string StandardError { get; set; }
+
+        /// <summary>
+        /// Tool set ran by command.
+        /// </summary>
+        public string ToolSet { get; set; }
```
brdeyo, line 41: Rename: Let's use the term "ToolName" for this as it is consistent with nomenclature we use elsewhere as well as in the Metrics data table.

yangpanMS:
Now we have a release snapshotted, please update this PR with the changes you want to check in

src/VirtualClient/VirtualClient.Actions/CoreMark/CoreMarkExecutor.cs
```diff
@@ -108,17 +108,22 @@ private async Task ExecuteWorkloadAsync(string pathToExe, string commandLineArgu
             {
                 if (!cancellationToken.IsCancellationRequested)
                 {
+                    process.ProcessDetails.ToolSet = "CoreMark";
                     if (process.IsErrored())
                     {
-                        await this.LogProcessDetailsAsync(process, telemetryContext, "CoreMark", logToFile: true);
+                        await this.LogProcessDetailsAsync(process.ProcessDetails, telemetryContext, logToFile: true);
```
brdeyo, line 114: I would like to revisit this implementation. I think that we can probably minimize the number of changes here by placing the point at which we reference the ProcessDetails abstraction within an extension method so we can still use the IProcessProxy as is. We will add in an extension method to match for the ISshclientProxy that takes in whatever parameters it needs to take in. The 3rd extension method would then take in a ProcessDetails object and would be called by the other 2 here underneath.

yangpanMS:
@deep1712 could we close this?

yangpanMS:
Closing since the sshclient is merged and otehr changes are not needed

PR #87, opened 2023-04-21
Feature to Install NVIDIA GPU Drivers and CUDA Toolkit in Windows

imadityaa:
This PR adds the feature to Install NVIDIA GPU Drivers and CUDA Toolkit in Windows:
The User can either provide the Web URL to download the drivers or download it from a blob storage using DependencyPackageInstallation followed by Installing the Drivers.


src/VirtualClient/VirtualClient.Dependencies/CUDAAndNvidiaGPUDriverInstallation.cs
```diff
@@ -105,6 +122,41 @@ public string Username
...
+            {
+                return this.Parameters.GetValue<string>(nameof(CudaAndNvidiaGPUDriverInstallation.WinCommandLineArgs), "-y -s");
+            }
+        }
+
+        /// <summary>
+        /// If this is true, VC expects CUDA and NVIDIA Drivers Installer Package from Blob Storage.
+        /// If false, CUDA and NVIDIA Drivers installer will be downloaded from Web based on link provided.
+        /// </summary>
+        public bool WinCudaDriversPackageDownloadedFromBlob
```
yangpanMS, line 140: Do we need this? I'm not seeing this in the profile

website/docs/workloads/mlperf/mlperf-profiles.md
```diff
@@ -33,9 +33,9 @@ Runs the MLPerf benchmark workload to test GPU performance.
   |-----------------------|---------|---------------|
   | Username     | Optional. User which needs to be created in container to run MLPerf benchmarks. | testuser  |
   | DiskFilter     | Optional. Filter to decide the disk that will be used to download benchmarks. Since benchmarks data is around 800gb, default disk filter is greater than 1000gb. | SizeGreaterThan:1000gb  |
-  | CudaVersion     | Optional. Version of CUDA that needs to be installed. | 11.6  |
-  | DriverVersion     | Optional. Version of GPU driver that needs to be installed. | 510  |
-  | LocalRunFile     | Optional. Link to download specified CUDA and GPU driver versions. | https://developer.download.nvidia.com/compute/cuda/12.0.0/local_installers/cuda_12.0.0_525.60.13_linux.run  |
+  | LinuxCudaVersion     | Optional. Version of CUDA that needs to be installed. | 11.6  |
```
yangpanMS, line 36: I know we didn't have one document for the Linux driver installation, but could you help start one in website/docs/dependencies for this dependency class? You can add the Windows one first

src/VirtualClient/VirtualClient.Dependencies/CUDAAndNvidiaGPUDriverInstallation.cs
```diff
@@ -156,13 +208,23 @@ await this.stateManager.SaveStateAsync(nameof(CudaAndNvidiaGPUDriverInstallation
...
+
+                    await this.stateManager.SaveStateAsync(nameof(this.CudaAndNvidiaGPUDriverInstallationOnWindowsAsync), new State(), cancellationToken)
+                        .ConfigureAwait(false);
+
+                    SystemManagement.IsRebootRequested = this.RebootRequired;
+                }
                 else
                 {
                     // CUDA and Nvidia driver installation for other platforms to be added.
                     throw new WorkloadException(
```
yangpanMS, line 224: Just take this part out, right now under .NET 6 there is only Unix and Windows

yangpanMS (commented):
I'm not seeing an example of how we are going to install it in Windows. Please add where did you get the driverInstall.exe, and how we versioned it etc. Ideally we should be getting it with wget, and with custom versions if possible.

src/VirtualClient/VirtualClient.Dependencies/CUDAAndNvidiaGPUDriverInstallation.cs
imadityaa: There are two ways by which the driver download will be supported, one is through the Web (which I have put in the profile), and the other is blob. The blob part should prove helpful in either disconnected scenarios or when GPU vendors share private drivers that are to be used for NPI skus, they can be uploaded in blob and used. I'll add the use-case in the documentation for this dependency.

website/docs/workloads/mlperf/mlperf-profiles.md
imadityaa: Added the documentation for Windows as well as Linux.

src/VirtualClient/VirtualClient.Dependencies/CUDAAndNvidiaGPUDriverInstallation.cs
imadityaa: Modified as suggested.

imadityaa:
> I'm not seeing an example of how we are going to install it in Windows. Please add where did you get the driverInstall.exe, and how we versioned it etc. Ideally we should be getting it with wget, and with custom versions if possible.

Added examples in the documentation. Modified the code to use wget, earlier it was being downloaded using HttpClient.

.pipelines/azure-pipelines.yml
```diff
@@ -87,4 +87,4 @@ steps:
     MaxRetryAttempts: '5'
 
 - script: $(Build.SourcesDirectory)\upload-packages.cmd $(Build.SourcesDirectory)\out\packages $(NUGETORGAPIKEY)
-  displayName: 'Publish NuGet Packages'
+  displayName: 'Publish NuGet Packages'
```
yangpanMS, line 90: Did you make any whitespace changes here?

website/docs/dependencies/0031-install-cuda-and-nvidia-drivers.md
```diff
@@ -0,0 +1,111 @@
...
+### Linux example for downloading drivers
+A sample URL for NVIDIA Drivers RunFile for Linux Ubuntu is mentioned in example. The exact URL for the specific OS and Driver Version can be taken from CUDA Toolkit website, given above.
+  <div class="code-section">
+
+  ```json
+  {
+      "Type": "NvidiaCudaInstallation",
+      "Parameters": {
+          "Scenario": "InstallNvidiaCuda",
+          "CudaVersion": "12.0",
```
yangpanMS, line 104: Please update the parameters here with the new name

website/docs/dependencies/0031-install-cuda-and-nvidia-drivers.md
```diff
@@ -0,0 +1,111 @@
...
+          "Extract": true
+      }
+  },
+  {
+      "Type": "CudaAndNvidiaGPUDriverInstallation",
+      "Parameters": {
+          "Scenario": "InstallCudaAndNvidiaGPUDriverForWindows",
+          "WinCudaDriversPackageDownloadedFromBlob": true,
+          "RebootRequired": false,
+          "WinCommandLineArgs": "-y -s",
```
yangpanMS, line 87: -y -s seems pretty standard, would there be any other reasonable options? If not let's remove this parameter

website/docs/dependencies/0031-install-cuda-and-nvidia-drivers.md
```diff
@@ -0,0 +1,111 @@
...
+
+| **Parameter** | **Required** | **Description**            |                 **Default**                     |
+|---------------|--------------|----------------------------|-------------------------------------------------|
+| PackageName   | Yes          | The logical name of the package that will be registered with the Virtual Client runtime. This name can be used by other profile components to reference the installation parent directory location for Drivers. |  |
+| Scenario      | No           | A name/identifier for the specific component in the profile. This is used for telemetry purposes only with components in dependency sections of the profile (i.e. cannot be used with --scenarios option on the command line). |  |
+| WinCommandLineArgs | No | The command line arguments that will be used with the Windows exe installer. | -y -s |
+| WinCudaDriversPackageDownloadedFromBlob | No | A 'true' value would mean that drivers package is downloaded from Blob and DependencyPackageInstallation is present in dependencies. A 'false' value would imply that WgetPackageInstallation is present in dependencies to download the NVIDIA drivers installer from Web | false |
```
yangpanMS, line 21: Is there a way we can handle this in the code? Regardless of we used wget or blobPackage download to download package, we should still be able to find the packages via package name.

src/VirtualClient/VirtualClient.Dependencies/Packaging/WgetPackageInstallation.cs
```diff
@@ -149,6 +149,15 @@ protected override async Task ExecuteAsync(EventContext telemetryContext, Cancel
 
                                     await this.packageManager.ExtractPackageAsync(downloadedPackagePath, installationPath, cancellationToken, archiveType);
                                 }
+                                else
+                                {
+                                    // This section is for the standalone files which are not zipped. The file will be copied to a location in installationPath and then
+                                    // file at original download path is deleted as in above case for zip file.
+                                    string installationFilePath = this.GetPackagePath(this.PackageName, Path.GetFileName(this.PackageUri.ToString()));
```
yangpanMS, line 156: Could we add a unit test for the path and copy for this scenario? Also let's not imply a file downloaded is an installFilePath. In addition, in comment at line 155 you mentioned the file at original path would be deleted? In this case you are not deleting it?

website/docs/dependencies/0031-install-cuda-and-nvidia-drivers.md
imadityaa: updated

website/docs/dependencies/0031-install-cuda-and-nvidia-drivers.md
imadityaa: Handled in code as suggested, and removed the parameter.

src/VirtualClient/VirtualClient.Dependencies/Packaging/WgetPackageInstallation.cs
imadityaa: Added UnitTests for this part as suggested.
Modified the variable name for better understanding.
The original downloaded file is being deleted at the end of function at line # 177

website/docs/dependencies/0031-install-cuda-and-nvidia-drivers.md
imadityaa: removed the parameter

.pipelines/azure-pipelines.yml
imadityaa: It happened due to changed line ending. Resolved it with notepad :)

src/VirtualClient/VirtualClient.Dependencies/CUDAAndNvidiaGPUDriverInstallation.cs
```diff
@@ -38,53 +43,65 @@ public CudaAndNvidiaGPUDriverInstallation(IServiceCollection dependencies, IDict
...
 
             set
             {
-                this.Parameters[nameof(CudaAndNvidiaGPUDriverInstallation.LocalRunFile)] = value;
+                this.Parameters[nameof(CudaAndNvidiaGPUDriverInstallation.LinuxLocalRunFile)] = value;
+            }
+        }
+
+        /// <summary>
+        /// Determines whether Reboot is required or not after Driver installation
```
yangpanMS, line 98: This now only applies to Windows, if Windows never needs a reboot, let's just remove this parameter. When appearing on the profile this will confuse user

yangpanMS (approved):
Looks good with all the clean up. I would like to see the RebootRequired removed if possible, other wise we need to make it clear that Linux will reboot anyway, that parameter only applies to Windpws

yangpanMS (approved):

src/VirtualClient/VirtualClient.Dependencies/CUDAAndNvidiaGPUDriverInstallation.cs
imadityaa: resolving as per discussion

PR #103, opened 2023-05-30
Adding bufferWait dependency

nmalkapuram:
Enables a step to wait for given time from profile of any workload.

### DIFF
--- a/src/VirtualClient/VirtualClient.Dependencies.UnitTests/BufferTimeWaiterTests.cs
+++ b/src/VirtualClient/VirtualClient.Dependencies.UnitTests/BufferTimeWaiterTests.cs
@@ -1,0 +1,68 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace VirtualClient.Dependencies
+{
+    using System;
+    using System.Runtime.InteropServices;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Moq;
+    using NUnit.Framework;
+    using VirtualClient.Common.Telemetry;
+    using VirtualClient.Contracts;
+
+    [TestFixture]
+    [Category("Unit")]
+    internal class BufferTimeWaiterTests
+    {
+        private MockFixture mockFixture;
+
+        public void SetupDefaults(PlatformID platform, Architecture architecture = Architecture.X64)
+        {
+            this.mockFixture = new MockFixture();
+            this.mockFixture.Setup(platform, architecture);
+        }
+
+        [Test]
+        [TestCase(PlatformID.Unix, Architecture.X64)]
+        public async Task BufferTimeWaiterWaitsForExpectedAmountOfTime(PlatformID platform, Architecture architecture)
+        {
+            this.SetupDefaults(platform, architecture);
+            this.mockFixture.Parameters[nameof(BufferTimeWaiter.BufferTimeInSec)] = 1;
+
+            using (TestBufferTimeWaiter testBufferTimeWaiter = new TestBufferTimeWaiter(this.mockFixture))
+            {
+                DateTime startTime = DateTime.Now;
+                await testBufferTimeWaiter.ExecuteAsync(CancellationToken.None).ConfigureAwait(false);
+                DateTime endTime = DateTime.Now;
+                if ((endTime - startTime).TotalSeconds > long.Parse(this.mockFixture.Parameters[nameof(BufferTimeWaiter.BufferTimeInSec)].ToString()))
+                {
+                    Assert.IsTrue(true);
+                }
+                else
+                {
+                    Assert.IsTrue(false);
+                }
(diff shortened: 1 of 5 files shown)


src/VirtualClient/VirtualClient.Dependencies/BufferTimeWaiter.cs
```diff
@@ -0,0 +1,56 @@
...
+    using Polly;
+    using VirtualClient.Common;
+    using VirtualClient.Common.Extensions;
+    using VirtualClient.Common.Telemetry;
+    using VirtualClient.Contracts;
+
+    /// <summary>
+    /// Provides functionality to wait for given time..
+    /// </summary>
+    public class BufferTimeWaiter : VirtualClientComponent
```
yangpanMS, line 20: Thanks for adding this. Let's just call it WaitExecutor, and let's add a minimum amount of document for this.

src/VirtualClient/VirtualClient.Dependencies/BufferTimeWaiter.cs
```diff
@@ -0,0 +1,56 @@
...
+// Licensed under the MIT License.
+
+namespace VirtualClient.Dependencies
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Extensions.DependencyInjection;
+    using Polly;
```
yangpanMS, line 11: unnecessary reference

src/VirtualClient/VirtualClient.Dependencies/BufferTimeWaiter.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        /// <summary>
+        /// The time in seconds to wait.
+        /// </summary>
+        public int BufferTimeInSec
+        {
+            get
+            {
+                return this.Parameters.GetValue<int>(nameof(this.BufferTimeInSec), 2);
```
yangpanMS, line 39: Could we do timespan like how we do everything else? so that you can even define milliseconds

src/VirtualClient/VirtualClient.Dependencies.UnitTests/BufferTimeWaiterTests.cs
```diff
@@ -0,0 +1,68 @@
...
+            this.mockFixture = new MockFixture();
+            this.mockFixture.Setup(platform, architecture);
+        }
+
+        [Test]
+        [TestCase(PlatformID.Unix, Architecture.X64)]
+        public async Task BufferTimeWaiterWaitsForExpectedAmountOfTime(PlatformID platform, Architecture architecture)
+        {
+            this.SetupDefaults(platform, architecture);
+            this.mockFixture.Parameters[nameof(BufferTimeWaiter.BufferTimeInSec)] = 1;
```
yangpanMS, line 32: Once you change the wait duration to milliseconds, you can change this test to wait for only 50ms and verify that, makes the test run a lot quicker

src/VirtualClient/VirtualClient.Dependencies/WaitExecutor.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <param name="parameters">A series of key value pairs that dictate runtime execution.</param>
+        public WaitExecutor(IServiceCollection dependencies, IDictionary<string, IConvertible> parameters)
+            : base(dependencies, parameters)
+        { 
+        }
+
+        /// <summary>
+        /// The performance snapshot interval defined in the profile arguments.
+        /// </summary>
+        public TimeSpan TimeInterval
```
yangpanMS, line 35: Sorry about this last time, I would prefer calling this Duration.

website/docs/dependencies/0002-wait-executor.md
```diff
@@ -0,0 +1,86 @@
...
+        "CompilerVersion": "$.Parameters.CompilerVersion",
+        "SpecProfile": "intrate",
+        "PackageName": "speccpu2017",
+        "RunPeak": "$.Parameters.RunPeak",
+        "BaseOptimizingFlags": "$.Parameters.BaseOptimizingFlags",
+        "PeakOptimizingFlags": "$.Parameters.PeakOptimizingFlags"
+      }
+    },
+    {
+      "Type": "WaitExecutor",
```
yangpanMS, line 56: Maybe we can show we can use this in both Action and Dependency? Have you tested it?

yangpanMS (approved):

website/docs/dependencies/0002-wait-executor.md
nmalkapuram: Have tested in actions, worked well. 
I believe it should also work in dependencies. Will test it.

yangpanMS (approved):
Excellent, thank you!

PR #106, opened 2023-06-05
fixing single role layout bug

ericavella:
creating check in mysql configuration to only check layout if we are in multi role


yangpanMS (approved):
Just make sure you also test back at the server/client

yangpanMS (approved):

PR #111, opened 2023-06-15
Feature to have metrics in csv format on system

imadityaa:
The PR adds the feature in Virtual Client to retain Workload metrics in the system in CSV Format for users to refer to and use.

### DIFF
--- a/src/VirtualClient/VirtualClient.Common/Telemetry/Serilog/SerilogCsvFileLoggerProvider.cs
+++ b/src/VirtualClient/VirtualClient.Common/Telemetry/Serilog/SerilogCsvFileLoggerProvider.cs
@@ -1,0 +1,65 @@
+﻿// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+namespace VirtualClient.Common.Telemetry
+{
+    using System;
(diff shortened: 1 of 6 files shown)


src/VirtualClient/VirtualClient.Common/Telemetry/Serilog/SerilogCsvFileLoggerProvider.cs
```diff
@@ -0,0 +1,65 @@
...
+    using ILogger = Microsoft.Extensions.Logging.ILogger;
+
+    /// <summary>
+    /// Provides methods for creating <see cref="ILogger"/> instances that can
+    /// be used to log events/messages to a local csv file.
+    /// </summary>
+    [LoggerSpecialization(Name = SpecializationConstant.StructuredLogging)]
+    public sealed class SerilogCsvFileLoggerProvider : ILoggerProvider
+    {
+        private LoggerConfiguration configuration;
```
brdeyo, line 20: Not a blocker, but you do not need to have extra line breaks after local member variables. It does not improve readability for the most part.

brdeyo (approved):

PR #131, opened 2023-07-10
Enabling redis with TLS feature

nmalkapuram:


website/docs/workloads/redis/redis-profiles.md
```diff
@@ -50,11 +50,14 @@ idea. The name of the client must match the name of the system or the value of t
 ## PERF-REDIS.json
 Runs the Memtier workload against to generate various network traffic patterns against a Redis server. Although this is the default client workload
 the Redis benchmark itself can also be run to evaluate the performance of the Redis server.
+We have two profiles for Redis.One supports redis with TLS and one without TLS.
+
+* [Workload Profile without TLS](https://github.com/microsoft/VirtualClient/blob/main/src/VirtualClient/VirtualClient.Main/profiles/PERF-REDIS.json) 
+* [Workload Profile with TLS](https://github.com/microsoft/VirtualClient/blob/main/src/VirtualClient/VirtualClient.Main/profiles/PERF-REDIS-TLS.json)
```
yangpanMS, line 56: Do we have any doc on how to generate the .crt or refresh it? Does the crt has an expiration?

src/VirtualClient/VirtualClient.Main/profiles/PERF-REDIS-TLS.json
```diff
@@ -0,0 +1,226 @@
...
+      "ClientInstances": 1,
+      "Duration": "00:02:00"
+    },
+    "Actions": [
+        {
+            "Type": "RedisServerExecutor",
+          "Parameters": {
+            "Scenario": "redis_server_tls",
+            "PackageName": "redis",
+            "CommandLine": "--tls-cert-file {ScriptPath:redis}/RedisResources/redis.crt   --tls-key-file {ScriptPath:redis}/RedisResources/redis.key --tls-ca-cert-file {ScriptPath:redis}/RedisResources/ca.crt --protected-mode no --io-threads {ServerThreadCount} --maxmemory-policy noeviction --ignore-warnings ARM64-COW-BUG --save",
```
yangpanMS, line 22: Can we have all these in the code, if we already have the boolean option? It will confuse use as they might think having the boolean already works

nmalkapuram: That's a valid point. That's one of the reasons why I haven't added IsTLSEnabled option as user changeable parameter. The confusion arises when user creates a custom profile.

website/docs/workloads/redis/redis-profiles.md
nmalkapuram: We have a script file In Redis that generates these certificates and keys. It's part of redis binaries. As of I know we don't have expiration for these crt and keys. Rather than generating crt and keys everytime we run the workload, we can generate them once and can use those. 
In Redis folder we have the script file at redis/utils/gen-test-certs.sh which generates crt and keys at redis/tests/tls folder.

yangpanMS: So by putting those keys in source, does that mean anyone can get the key we generate and decrypt the redis messages? Which might be fine because this is benchmarking with fake data, but we might want to make it clear in the document

src/VirtualClient/VirtualClient.Main/profiles/PERF-REDIS-TLS.json
yangpanMS: I would vote for just the boolean in parameter, and inject the arguments in the code

website/docs/workloads/redis/redis-profiles.md
nmalkapuram: Correct. Here we are not concerned about the security aspect but just the benchmarking results. Sure will update the document to make it more clear. Thanks Yang

yangpanMS (approved):

PR #133, opened 2023-07-12
Feature to add a VCComponent parameter ContentPathTemplate to support custom file upload blob Path

imadityaa:


src/VirtualClient/VirtualClient.Monitors/FileUploadMonitor.cs
```diff
@@ -114,6 +114,7 @@ private async Task ProcessFileUploadsAsync(IBlobManager blobManager, EventContex
                 catch (OperationCanceledException)
                 {
                     // Expected whenever ctrl-C is used. Do a check once more, without cancellationToken and break;
+                    await this.UploadFilesAsync(blobManager, telemetryContext);
```
imadityaa, line 117: This resolves an issue where the last set of logs for a given VC run were not getting uploaded on blob

src/VirtualClient/VirtualClient.Core/ProfileExecutor.cs
```diff
@@ -708,6 +707,11 @@ private static void OutputComponentStart(string componentType, VirtualClientComp
                             runtimeComponent.Metadata.AddRange(this.Metadata, true);
                         }
 
+                        if (this.Profile.Parameters?.Any() == true)
+                        {
+                            runtimeComponent.Parameters.AddRange(this.Profile.Parameters, true);
```
imadityaa, line 712: Provide visibility of all commandline parameters to profile

src/VirtualClient/VirtualClient.Monitors/FileUploadMonitor.cs
yangpanMS: Is it possible for us to get stuck here? if the upload takes forever? We should nevertheless put a hard stop to prevent VC fails to stop when there is a long backlog of files to upload

src/VirtualClient/VirtualClient.Core/ProfileExecutor.cs
yangpanMS: Is there a case the current implementation doesn't work? I'm worried that we might add parameters to say SPECjbb, to another workloads or monitor. Functionally it's fine but it bloats the telemetry

src/VirtualClient/VirtualClient.Contracts/VirtualClientComponent.cs
```diff
@@ -117,6 +117,18 @@ public string ContentPathFormat
...
 
+        /// <summary>
+        /// Parameter defines the content path format/structure using a template to use when uploading content
+        /// to target storage resources. When not defined the 'Default' structure is used.
+        /// </summary>
+        public string ContentPathTemplate
+        {
+            get
+            {
+                return this.Parameters.GetValue<string>(nameof(this.ContentPathTemplate), "{experimentId}/{agentId}/{toolName}/{role}/{scenario}");
```
yangpanMS, line 128: It's nice to see this!

yangpanMS (commented):
It is great feature to have! Thank you for making the change. Please modify the document we have on file uploading and the structure.

src/VirtualClient/VirtualClient.Core/ProfileExecutor.cs
imadityaa: From my understanding, the Profile.Parameters has the global parameters that are provided either through Profile parameters or from CommandLine. While, the vcComponent.Parameters has action/monitor level parameters. I modified the AddRange to not replace the keys in dictionary if already present, so any parameter defined in action would not be overwritten here by global parameter here.

src/VirtualClient/VirtualClient.Common/Extensions/CollectionExtensions.cs
```diff
@@ -42,7 +42,14 @@ public static void AddRange<T>(this IDictionary<string, T> dictionary, IEnumerab
                 }
                 else
                 {
-                    dictionary.Add(entry.Key, entry.Value);
+                    try
```
yangpanMS, line 45: As I mentioned in the other comment, we need to be a little careful about this. We don't want some class to have extra parameter that doesn't apply to it. Each dictionary should be separate, and if you are somehow merging and have conflicts, the caller should decide whether to ignore this error. 

Also imagine this profile:
One workload has a Command: "-abc"
The other workload has command: undefined (user wants default)
By allowing parameters of one workload flow to others, you now allow the first parameter to overwrite the next one

yangpanMS (approved):
As discussed, we could get this in now, but please help refactor to the better design that we discussed.

src/VirtualClient/VirtualClient.Common/Extensions/CollectionExtensions.cs
imadityaa: Added a Global Parameter

PR #136, opened 2023-07-19
Making the GlobalParameters Static for access to child classes in Client-server workloads

imadityaa:
This PR is for Client Server workloads. In Client Server workloads, a common executor calls the client and server class by declaring objects and just passing the dependencies and component Parameters. So, the client/server executors do not have visibility to GlobalParameters.

Making the GlobalParameters a static component as they are expected to be common across all components (unlike component level parameters)


yangpanMS (approved):
I will approve to unblock you, but as discussed, please keep those changes footprint as small as possible, as we will refactor these. Also let me know if this breaks the extension repos

PR #146, opened 2023-08-01
Users/nmalkapuram/redis tls combined profile

nmalkapuram:
Combining redis TLS, without TLS profile.
Adding redis installation from apt package manager dependency.


src/VirtualClient/VirtualClient.Actions.UnitTests/Redis/RedisServerExecutorTests.cs
```diff
@@ -38,7 +38,7 @@ public void SetupTests()
                 ["Port"] = 6379,
                 ["ServerInstances"] = 1,
                 ["ServerThreadCount"] = 4,
-                ["IsTLSEnabled"] = false
+                ["IsTLSEnabled"] = "no"
```
brdeyo, line 41: Let's use the original boolean value vs. a string literal boolean. The property in your component can use the TryGetValue() method/technique for checking to see if the parameter is there and returning a default value if not.

src/VirtualClient/VirtualClient.Actions/Memtier/MemtierBenchmarkClientExecutor.cs
```diff
@@ -311,8 +346,15 @@ private Task ExecuteWorkloadsAsync(IPAddress serverIPAddress, ServerState server
                             // memtier_benchmark Documentation:
                             // https://github.com/RedisLabs/memtier_benchmark
 
-                            string commandArguments = commandArguments = $"--server {serverIPAddress} --port {serverPort} {this.CommandLine}";
-                            
+                            if (string.Equals(this.IsTLSEnabled, "yes", StringComparison.OrdinalIgnoreCase))
```
brdeyo, line 349: As mentioned above, let's go back to IsTLSEnabled being a proper boolean data type.

src/VirtualClient/VirtualClient.Actions.UnitTests/Redis/RedisServerExecutorTests.cs
nmalkapuram: In the same profile we are using this value of IsTLSEnabled in the dependencies 
{
      "Type": "ExecuteCommand",
      "Parameters": {
        "Scenario": "CompileRedis",
        "SupportedPlatforms": "linux-x64,linux-arm64",
        "BUILD_TLS": "$.Parameters.IsTLSEnabled",
        "Command": "make BUILD_TLS={BUILD_TLS}",
        "WorkingDirectory": "{PackagePath:redis}"
      }
    }
BUILD_TLS takes "yes" or "no values to build Redis with TLS and without TLS respectively. To have this dependency as part of profile I was using IsEnabledTLS as string rather than as boolean.

If we could have ternary operator feature added in the profile expression evaluator, we can add ISEnabledTLS as true and just derive yes from this ISEnabledTLS value in the profile itself.

brdeyo (approved):

src/VirtualClient/VirtualClient.Dependencies/RedisPackageInstallation.cs
```diff
@@ -0,0 +1,158 @@
...
+    using VirtualClient.Common.Extensions;
+    using VirtualClient.Common.Platform;
+    using VirtualClient.Common.Telemetry;
+    using VirtualClient.Contracts;
+
+    /// <summary>
+    /// Provides functionality for installing latest version of Redis from Apt package manager on specific OS distribution version.
+    /// </summary>
+    [UnixCompatible]
+    public class RedisPackageInstallation : VirtualClientComponent
```
brdeyo, line 26: Why do we need a RedisPackageInstallation to install redis from "apt" package stores? At a minimum, it seems like we should derive from LinuxPackageInstallation and simply override a few methods to keep the duplication to a minimum. This doesn't need to block this PR, but I'd like to think a bit more on propagating new dependency implementations before we figure out how to genericize the need. We might just need to extend LinuxPackageInstallation to have a few more features if it does not currently support the needs here.

PR #147, opened 2023-08-04
Users/prachisingla/add monitor iterations in base monitor

psingla1210:


src/VirtualClient/VirtualClient.Contracts/VirtualClientIntervalBasedMonitor.cs
```diff
@@ -14,10 +14,11 @@ namespace VirtualClient.Contracts
     public abstract class VirtualClientIntervalBasedMonitor : VirtualClientComponent
     {
         /// <summary>
-        /// The default frequency that monitors run at (every one minute)
+        /// The default frequency that monitors run at (every five minutes)
```
yangpanMS, line 17: Thank you!

src/VirtualClient/VirtualClient.Contracts/VirtualClientIntervalBasedMonitor.cs
```diff
@@ -14,10 +14,11 @@ namespace VirtualClient.Contracts
...
         /// </summary>
         private static TimeSpan defaultFrequency = TimeSpan.FromMinutes(5);
         private static TimeSpan defaultWarmupPeriod = TimeSpan.FromMinutes(5);
+        private static long defaultIterations = 6315489358112;
```
yangpanMS, line 21: You could use something like long.maxvalue

src/VirtualClient/VirtualClient.Contracts/VirtualClientIntervalBasedMonitor.cs
```diff
@@ -68,5 +69,17 @@ public TimeSpan MonitorWarmupPeriod
                     nameof(VirtualClientIntervalBasedMonitor.MonitorWarmupPeriod), VirtualClientIntervalBasedMonitor.defaultWarmupPeriod);
             }
         }
+
+        /// <summary>
+        /// The number of iterations with which this monitor runs
+        /// </summary>
+        public long MonitorIterations
```
yangpanMS, line 76: What's the reason to put a maximum when user doesn't define it? User might want infinite monitor iterations, like SAP

yangpanMS (approved):

src/VirtualClient/VirtualClient.Contracts/VirtualClientIntervalBasedMonitor.cs
psingla1210: I have set default value, reason behind it is if let say a monitor is using this to determine number of iterations and user removed the line which is providing this value in profile file then to protect the code from crashing, i gave a very high value which is like infinite only, do you see any concern?

yangpanMS (approved):

PR #151, opened 2023-08-16
Users/nmalkapuram/furmark test on juno

nmalkapuram:
Updating documentation
Checked for Furmark run through Juno experiments.


yangpanMS (approved):
Please note Juno is internal and let's not put juno in any of VC's code or documents.

PR #153, opened 2023-08-17
Adding a new VirtualClient parameter for contentPathPattern

imadityaa:


src/VirtualClient/VirtualClient.Monitors/FileUploadMonitor.cs
```diff
@@ -91,23 +91,20 @@ private async Task ProcessFileUploadsAsync(IBlobManager blobManager, EventContex
 
                 try
                 {
-                    await this.Logger.LogMessageAsync($"{this.TypeName}.ProcessFileUploads", relatedContext, async () =>
```
imadityaa, line 94: Removed the logger here and added it in successive block to avoid the flooding up of traces with FileUploadMonitor log every few seconds.

src/VirtualClient/VirtualClient.Contracts/VirtualClientComponent.cs
```diff
@@ -121,13 +121,7 @@ public string ContentPathFormat
         /// Parameter defines the content path format/structure using a template to use when uploading content
         /// to target storage resources. When not defined the 'Default' structure is used.
```
yangpanMS, line 122: In the comment you mentioned default, it would be good to also add default to comment if it's not in the code

src/VirtualClient/VirtualClient.Main/OptionFactory.cs
```diff
@@ -121,6 +121,27 @@ public static Option CreateContentStoreOption(bool required = true, object defau
...
 
+        /// <summary>
+        /// Command line option defines the template for virtual folder structure to be followed in
+        /// contentstore when logs are uploaded to it.
+        /// </summary>
+        /// <param name="required">Sets this option as required.</param>
+        /// <param name="defaultValue">Sets the default value when none is provided.</param>
+        public static Option CreateContentPathPatternOption(bool required = true, object defaultValue = null)
+        {
+            Option<string> option = new Option<string>(new string[] { "--contentPathPattern", "--contentpathpattern", "--contentPath", "--cspt" })
```
yangpanMS, line 132: hmm what is cspt acronym of?

src/VirtualClient/VirtualClient.Main/Program.cs
```diff
@@ -192,6 +192,9 @@ internal static CommandLineBuilder SetupCommandLine(string[] args, CancellationT
                 // --contentStore
                 OptionFactory.CreateContentStoreOption(required: false),
 
+                // --contentPathPattern
+                OptionFactory.CreateContentPathPatternOption(required: false, "{experimentId}/{agentId}/{toolName}/{role}/{scenario}"),
```
yangpanMS, line 196: I would rather this default pattern set later in the code, not all the way front in Program.

yangpanMS (approved):
Looks good! Please check the comments

src/VirtualClient/VirtualClient.Main/Program.cs
imadityaa: Modified as suggested. Now the default is set in VCComponent

yangpanMS (approved):
Everything is good, except, what is cspt? where does the s come from?

PR #155, opened 2023-08-17
Users/nmalkapuram/linpack with arm perf libraries

nmalkapuram:


src/VirtualClient/VirtualClient.Actions/Memcached/MemcachedExecutor.cs
```diff
@@ -58,7 +58,7 @@ public string Username
         {
             get
             {
-                string username = this.Parameters.GetValue<string>(nameof(HPLinpackExecutor.Username), string.Empty);
+                string username = this.Parameters.GetValue<string>(nameof(MemcachedExecutor.Username), string.Empty);
```
yangpanMS, line 61: thanks

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -260,6 +287,23 @@ private void ValidateParameters()
             }
         }
 
+        private async Task ConfigurePerformanceLibrary(EventContext telemetryContext, CancellationToken cancellationToken)
+        {
+            if (this.UsePerformanceLibraries)
```
yangpanMS, line 292: the two condition could be && together to make the code less indented

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
```diff
@@ -39,19 +48,22 @@
...
     },
     {
       "Type": "DependencyPackageInstallation",
       "Parameters": {
-        "Scenario": "InstallHPLinpackPackage",
+        "Scenario": "DownloadHPLPerformanceLibrariesPackage",
         "BlobContainer": "packages",
-        "BlobName": "hpl.2.3.zip",
-        "PackageName": "hplinpack",
+        "BlobName": "HPLPerformanceLibraries.1.0.0.zip",
```
yangpanMS, line 65: Let's keep the whole package name lower cased

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -107,7 +100,7 @@ public int NumberOfProcesses
         {
             get
             {
-                return this.Parameters.GetValue<int>(nameof(HPLinpackExecutor.NumberOfProcesses), Environment.ProcessorCount);
+                return this.Parameters.GetValue<int>(nameof(this.NumberOfProcesses), Environment.ProcessorCount);
```
yangpanMS, line 103: Please get these information from cpuinfo, instead of Environment.xxx directly

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -84,7 +77,7 @@ public string ProblemSizeN
         {
             get
             {
-                return this.Parameters.GetValue<string>(nameof(HPLinpackExecutor.ProblemSizeN), Environment.ProcessorCount * 10000);
+                return this.Parameters.GetValue<string>(nameof(this.ProblemSizeN), Environment.ProcessorCount * 1000);
```
yangpanMS, line 80: Is this a breaking change with previous releases? Meaning if user switch version, the number they are getting is different

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -45,35 +49,24 @@ public HPLinpackExecutor(IServiceCollection dependencies, IDictionary<string, IC
         }
 
         /// <summary>
-        /// The user who has the ssh identity registered for.
+        /// True if Hyperthreading is on
         /// </summary>
-        public string Username
+        public bool HyperThreadingOn
```
yangpanMS, line 54: Can we get this from lscpu? instead of user defined?

src/VirtualClient/VirtualClient.Actions.UnitTests/HPLinpack/HPLinpackExecutorTests.cs
```diff
@@ -92,14 +92,51 @@ public async Task HPLinpackExecutorExecutesWorkloadAsExpectedOnUbuntuPlatform(Pl
...
+        public async Task HPLinpackExecutorExecutesWorkloadAsExpectedWithPerformanceLibrariesOnUbuntuArm64Platform(PlatformID platform, Architecture architecture)
+        {
+            this.SetupDefaultMockBehavior(platform, architecture);
+            this.fixture.Parameters["UsePerformanceLibraries"] = true;
+
+            using (TestHPLExecutor executor = new TestHPLExecutor(this.fixture))
+            {
+                List<string> expectedCommands = new List<string>()
+                {
+                    $"sudo chmod +x {this.fixture.PlatformSpecifics.Combine(this.mockPath.Path, "ARM", "arm-performance-libraries_23.04.1_Ubuntu-22.04.sh")}",
```
yangpanMS, line 127: is it ubuntu22 specific?

yangpanMS (approved):
Looks good! Please add documents on some things user might have questions about. Like supported OS, ubuntu22 or ubuntu or all linux distros. Why we pick this problem size, should user use arm perf lib by default etc

src/VirtualClient/VirtualClient.Actions.FunctionalTests/HPLinpackProfileTests.cs
```diff
@@ -111,9 +113,13 @@ public async Task HPLinpackWorkloadProfileExecutesTheExpectedWorkloadsOnUnixX64P
 
             using (ProfileExecutor executor = TestDependencies.CreateProfileExecutor(profile, this.mockFixture.Dependencies))
             {
-                await executor.ExecuteAsync(ProfileTiming.OneIteration(), CancellationToken.None).ConfigureAwait(false);
+                WorkloadException exception = Assert.ThrowsAsync<WorkloadException>(
+                    () => executor.ExecuteAsync(ProfileTiming.OneIteration(), CancellationToken.None));
+                // await executor.ExecuteAsync(ProfileTiming.OneIteration(), CancellationToken.None).ConfigureAwait(false);
```
psingla1210, line 118: Do we want to keep these comments?

src/VirtualClient/VirtualClient.Actions.FunctionalTests/HPLinpackProfileTests.cs
```diff
@@ -111,9 +113,13 @@ public async Task HPLinpackWorkloadProfileExecutesTheExpectedWorkloadsOnUnixX64P
...
 
-                WorkloadAssert.CommandsExecuted(this.mockFixture, expectedCommands.ToArray());
+                Assert.AreEqual(ErrorReason.PlatformNotSupported, exception.Reason);
+
+                // WorkloadAssert.CommandsExecuted(this.mockFixture, expectedCommands.ToArray());
```
psingla1210, line 122: this also?

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -159,12 +174,19 @@ protected override async Task InitializeAsync(EventContext telemetryContext, Can
...
 
             this.ValidateParameters();
-
             DependencyPath workloadPackage = await this.packageManager.GetPlatformSpecificPackageAsync(this.PackageName, this.Platform, this.CpuArchitecture, cancellationToken)
                                                     .ConfigureAwait(false);
 
             this.HPLDirectory = workloadPackage.Path;
+            
+            this.loggedInUserName = Environment.GetEnvironmentVariable("SUDO_USER");
+            if (this.loggedInUserName == null)
```
psingla1210, line 183: have we verified that it returns null if not run with sudo?

psingla1210: if we are not sure, i will suggest to check enviroment.username first and if it is sudo then go and check for SUDO_USER bt i will leave on you, how you want to go about it.

And i will also suggest if we can have a common function in vccomponentextensions to determine current user irrespective of linux and windows, you can choose not to and we can do later but if you can that will be very helpful.

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -260,6 +287,23 @@ private void ValidateParameters()
             }
         }
 
+        private async Task ConfigurePerformanceLibrary(EventContext telemetryContext, CancellationToken cancellationToken)
+        {
+            if (this.UsePerformanceLibraries)
+            {
+                if (this.CpuArchitecture == Architecture.Arm64)
+                {
+                    this.hplPerfLibraryInfo = "arm-performance-libraries_23.04.1_Ubuntu-22.04";
```
psingla1210, line 296: what will happen if someone will run on ubuntu 20.04 or any future ones?

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
```diff
@@ -6,30 +6,39 @@
...
   },
   "Parameters": {
-    "Username": "testuser",
     "CompilerName": "gcc",
     "CompilerVersion": "11",
     "ProblemSizeN": null,
     "BlockSizeNB": "256",
     "NumberOfProcesses": null,
-    "CCFLAGS": "-march=native"
+    "UsePerformanceLibraries": true,
```
psingla1210, line 14: Could you please also update documentation to reflect these parameters changes?

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
nmalkapuram: Nope even for the previous release, the default value of ProcessorCount * 10000 for large processor count is exceeding memory and we are facing memory issues.

src/VirtualClient/VirtualClient.Actions.UnitTests/HPLinpack/HPLinpackExecutorTests.cs
nmalkapuram: Yes. It is.

src/VirtualClient/VirtualClient.Actions.FunctionalTests/HPLinpackProfileTests.cs
nmalkapuram: Will clean this up.

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
nmalkapuram: > have we verified that it returns null if not run with sudo?

Yes, while running unittests which actually run without sudo, it was returning the value of Environment,Username.

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
psingla1210: If this type of change is only valid/tested for ubuntu22.04, could you please mention very clearly in our documentation that these options are only supported for this version and have not tested for other versions

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
yangpanMS: Does that mean linpack is not linear scaling? otherwise why for smaller processor it's working but not for larger ones? Azure has linear memory/cpu

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -84,7 +65,7 @@ public string ProblemSizeN
         {
             get
             {
-                return this.Parameters.GetValue<string>(nameof(HPLinpackExecutor.ProblemSizeN), Environment.ProcessorCount * 10000);
+                return this.Parameters.GetValue<string>(nameof(this.ProblemSizeN), this.cpuInfo.LogicalCoreCount * 5000);
```
yangpanMS, line 68: I see you just changed it to *5000, and the document says *1000. How did you choose this number?

yangpanMS (approved):

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -42,38 +45,17 @@ public HPLinpackExecutor(IServiceCollection dependencies, IDictionary<string, IC
...
-        {
-            get
-            {
-                return this.Parameters.GetValue<bool>(nameof(HPLinpackExecutor.HyperThreadingOn), true);
-            }
-
-            set
-            {
-                this.Parameters[nameof(HPLinpackExecutor.HyperThreadingOn)] = value;
+                return this.Parameters.GetValue<bool>(nameof(this.UsePerformanceLibraries), true);
```
yangpanMS, line 58: I think we discussed to make it false before we support it on Intel, AMD and ARM altogether

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -249,17 +269,61 @@ private void ThrowIfPlatformIsNotSupported()
                     $"'{PlatformSpecifics.LinuxX64}', '{PlatformSpecifics.LinuxArm64}'.",
                     ErrorReason.PlatformNotSupported);
             }
+            else if (this.Platform == PlatformID.Unix && this.CpuArchitecture != Architecture.Arm64 && this.UsePerformanceLibraries == true)
```
yangpanMS, line 272: See if you set the UsePerformanceLibrary to true by default. this profile is now default to crash on x64

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
```diff
@@ -6,30 +6,38 @@
     "SupportedOperatingSystems": "Ubuntu"
```
yangpanMS, line 6: Do we support all Ubuntu or just Ubuntu22?

nmalkapuram: Without performance libraries it is supported on all Ubuntu. But with performance library, only on Ubuntu 22. I have included this in the documentation.

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
nmalkapuram: Agree. I have changed the default value in the profile to false and forgot to change in the executor. Thanks for pointing this out.

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
yangpanMS: Then let's mention that in profile. you can put Ubuntu22 in the SupportedOperatingSystems

yangpanMS (approved):

yangpanMS (approved):

PR #163, opened 2023-08-28
Introduce the --fail-fast option for requesting VC to fail on first error regardless of severity.

brdeyo:
This currently applies to Actions in the profile only. Dependencies are ALWAYS fail-fast/immediately. Monitors are generally implemented to handle transient errors. We would need to change the implementation of Monitors to affect/modify this behavior.


src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-OPENSSL.json
```diff
@@ -11,7 +11,7 @@
             "Type": "OpenSslExecutor",
             "Parameters": {
                 "Scenario": "MD5",
-                "CommandArguments": "speed -elapsed -seconds 100 md5",
+                "CommandArguments": "speed -elapsed -seconds 100 md54321",
```
yangpanMS, line 14: Is this intentional?

yangpanMS (approved):

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-OPENSSL.json
brdeyo: No. Forgot to change it back but just did and pushed the change again. Good eyes!!

yangpanMS (approved):

PR #164, opened 2023-08-29
Users/prachisingla/adding parsers for ml perf

psingla1210:


website/docs/workloads/mlperf/mlperf-profiles.md
```diff
@@ -36,6 +36,7 @@ Runs the MLPerf benchmark workload to test GPU performance.
   | LinuxCudaVersion     | Optional. Version of CUDA that needs to be installed. | 11.6  |
   | LinuxDriverVersion     | Optional. Version of GPU driver that needs to be installed. | 510  |
   | LinuxLocalRunFile     | Optional. Link to download specified CUDA and GPU driver versions. | https://developer.download.nvidia.com/compute/cuda/12.0.0/local_installers/cuda_12.0.0_525.60.13_linux.run  |
+  | RequireCustomSystemSupport | Optional. This enables A100_PCIe_80GBx4 system support that was not supported by github repo of MLPerf. | true |
```
yangpanMS, line 39: Could you expand on this a bit more? It's not clear it
1. We only support A100*4
2. Only on A100 we need this
3. Only on A100*4 we need this

src/VirtualClient/VirtualClient.Actions/MLPerf/MLPerfMetricsParser.cs
```diff
@@ -14,6 +16,11 @@ namespace VirtualClient.Actions
     /// </summary>
     public class MLPerfMetricsParser : MetricsParser
     {
+        private static readonly string AccuracyResultsPattern = @"(\w+)\s*=\s*([\d.]+)"; 
+        private static readonly string PerformanceResultsPattern = @"([\w\d_.]+)\s*:\s*([\d.]+)";
```
yangpanMS, line 20: We have a Regex for double in our parser lib if you want to use it.

src/VirtualClient/VirtualClient.Actions/MLPerf/MLPerfExecutor.cs
```diff
@@ -220,8 +231,12 @@ await this.ThrowIfUnixDistroNotSupportedAsync(cancellationToken)
                 // add user in docker group and create scratch space
                 await this.ExecuteCommandAsync("usermod", $"-aG docker {this.Username}", this.NvidiaDirectory, cancellationToken);
                
-                // If GPUConfig is not included in the MLPerf code but is supported
-                this.ReplaceGPUConfigFilesToSupportAdditionalGPUs();
+                if (this.RequireCustomSystemSupport)
+                {
+                    // This enables A100_PCIe_80GBx4 system support that was not supported by github repo of MLPerf..
+                    this.ReplaceGPUConfigFilesToSupportAdditionalGPUs();
```
yangpanMS, line 237: If there is no harm in doing this, we can make it default

psingla1210: Considering further GPUs like H100 or later, these files are getting updated in original repo, so it is better that user use files directly from repo. We have added these for A100 to support additional systems along with what they support.

website/docs/workloads/mlperf/mlperf-profiles.md
psingla1210: Is it better now? We are providing additional this system support, original systems which open source support are supported already.

yangpanMS (approved):

PR #169, opened 2023-09-06
skip initialize optional parameter + tables/records auto check

ericavella:
for flexibility with users, adding an optional parameter to skip initialization of db
also bringing back a check for tables/records count stored in state
both are used to determine if db is set up; if one raises a flag, then db is set up
also adopting standards for duration parameter

### DIFF
--- a/src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPClientExecutor.cs
+++ b/src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPClientExecutor.cs
@@ -52,11 +52,11 @@ namespace VirtualClient.Actions
         /// <summary>
         /// The total time of execution option passed to Sysbench.
         /// </summary>
-        public string DurationSecs
+        public string Duration
         {
             get
             {
-                return this.Parameters.GetValue<string>(nameof(SysbenchOLTPClientExecutor.DurationSecs));
+                return this.Parameters.GetValue<string>(nameof(SysbenchOLTPClientExecutor.Duration));
             }
         }
 
@@ -89,7 +89,13 @@ namespace VirtualClient.Actions
         {
             get
             {
-                return this.Parameters.GetValue<string>(nameof(SysbenchOLTPClientExecutor.NumTables), 10);
+                this.Parameters.TryGetValue(nameof(SysbenchOLTPClientExecutor.NumTables), out IConvertible numTables);
+                return numTables?.ToString();
+            }
+
+            set
+            {
+                this.Parameters[nameof(this.NumTables)] = value;
             }
         }
 
@@ -130,11 +136,6 @@ namespace VirtualClient.Actions
             {
                 this.Parameters.TryGetValue(nameof(SysbenchOLTPClientExecutor.Threads), out IConvertible threads);
                 return threads?.ToString();
-            }
-
-            set
-            {
-                this.Parameters[nameof(this.Threads)] = value;
             }
         }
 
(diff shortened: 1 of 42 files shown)


src/VirtualClient/VirtualClient.Main/profiles/PERF-MYSQL-SYSBENCH-OLTP.json
```diff
@@ -27,7 +27,7 @@
                 "DatabaseName": "$.Parameters.DatabaseName",
                 "DatabaseScenario": "$.Parameters.DatabaseScenario",
                 "Role": "Client",
-                "DurationSecs": "1800",
+                "Duration": "00:01:00",
```
yangpanMS, line 30: Please update document for this parameter change

src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPClientExecutor.cs
```diff
@@ -335,24 +340,41 @@ private void SetupDatabaseParameters()
...
         }
 
         private async Task PrepareMySQLDatabase(EventContext telemetryContext, CancellationToken cancellationToken)
         {
             SysbenchOLTPState state = await this.stateManager.GetStateAsync<SysbenchOLTPState>(nameof(SysbenchOLTPState), cancellationToken);
 
-            if (!state.DatabaseInitialized)
+            int curTables = state.NumTables;
+            int curRecords = state.RecordCount;
+            int numTables = int.Parse(this.NumTables);
```
yangpanMS, line 374: Please make these properties just Int, int.parse shouldn't be needed

src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPClientExecutor.cs
```diff
@@ -335,24 +340,41 @@ private void SetupDatabaseParameters()
...
+            int recordCount = int.Parse(this.RecordCount);
+
+            if (!this.SkipInitialize || numTables > curTables || recordCount > curRecords)
```
yangpanMS, line 377: Could we still do what we discussed? Checking if current table configuration is same? instead of adding another SkipInitialize?

src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPExecutor.cs
```diff
@@ -293,16 +293,29 @@ public string DiskPathsArgument
                 }
             }
 
-            public bool DatabaseInitialized
+            public int NumTables
```
yangpanMS, line 296: Why we need SkipInitialie and NumTables to fix the same bug?

imadityaa: One Use Case Scenario: 

Let's say VC is supposed to be run multiple times on the SUT, so the MySQL Configuration will need to be performed only once, hence from 2nd run onwards, SkipInitialize can be set as true. Here' the prepare step will be needed if the Table Count/Record Count is more, and hence an "OR" condition is added there.

On the contrary, if SkipInitialize is set as false in the 2nd run, then Database will be reconfigured, hence the prepare step is needed here irrespective of TableCount/Size. So, the OR condition here helps to cover both scenarios.

ericavella: To add, this is an example workflow of consecutive VC runs on a SUT:
First Run
- MySQL db is set up, sysbench runs through, preserving a state of 100 records and 10 tables
Second Run
- MySQL db is deleted and recreated via Configuration dependency, but state preserved still refers to 100 records and 10 tables that are not there
- Adding skip initialize allows to initialize db despite state inconsistency. Also checking by records/tables allows to adjust the database between scenario executions

src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPClientExecutor.cs
```diff
@@ -85,45 +87,76 @@ public string Workload
...
+        public int RecordCount
         {
             get
             {
-                this.Parameters.TryGetValue(nameof(SysbenchOLTPClientExecutor.RecordCount), out IConvertible recordCount);
-                return recordCount?.ToString();
+                // default formulaic setup of the database
+                // records & threads depend on the core count
+
+                int coreCount = Environment.ProcessorCount;
```
yangpanMS, line 116: I recommend getting it using Get CpuInfo so that it could be mocked and tested. You are doing log and power calculation and I'd like to see that unit tested

src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPExecutor.cs
yangpanMS: Ok one more question, is there a reason for user to always re-initialize in 2nd run? Otherwise we assume true, and just judge by table/records count

imadityaa: By default, the profile has to support to initialize the database (because we will always need it in first run). 

From second run, it is surely better that user doesn't initialize, but it would mean assuming that user will change this parameter in profile/control through command line argument for subsequent runs, if the user doesn't do it, then he/she will face failure with the default profile recipe.

yangpanMS:
Could you help make sure the PR build passes?

yangpanMS (approved):
MySQL is pretty complex. Please verify it in brand new Ubuntu or using Juno to test it

yangpanMS (approved):

PR #171, opened 2023-09-13
Updating all thrrows of platform and arch to logging

psingla1210:
In Virtual Client, we used to throw error if workload is not supported for current platform and architecture but if we have multiple workloads in single profile and workloads are meant to run for different permutations and combinations of platform and architecture support, it makes sense that we go on with profile and not throw error if one of the intermediate workload is not supported.

### DIFF
--- a/src/VirtualClient/VirtualClient.Actions/SPECjvm/SpecJvmExecutor.cs
+++ b/src/VirtualClient/VirtualClient.Actions/SPECjvm/SpecJvmExecutor.cs
@@ -7,6 +7,7 @@ namespace VirtualClient.Actions
     using System.Collections.Generic;
     using System.IO;
     using System.IO.Abstractions;
+    using System.Runtime.InteropServices;
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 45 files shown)


src/VirtualClient/VirtualClient.Actions/SPECjvm/SpecJvmExecutor.cs
```diff
@@ -115,6 +116,24 @@ protected override async Task InitializeAsync(EventContext telemetryContext, Can
...
+        /// </summary>
+        protected override bool IsSupported()
+        {
+            bool isSupported = base.IsSupported()
+                && (this.Platform == PlatformID.Win32NT || this.Platform == PlatformID.Unix)
+                && (this.CpuArchitecture == Architecture.X64 || this.CpuArchitecture == Architecture.Arm64);
+
+            if (!isSupported)
+            {
+                this.Logger.LogNotSupported("SpecJvm", this.Platform, this.CpuArchitecture, EventContext.Persisted());
```
yangpanMS, line 131: maybe SPECjvm

yangpanMS (approved):

src/VirtualClient/VirtualClient.Actions/SPECjvm/SpecJvmExecutor.cs
psingla1210: somehow i am not able to push in this branch. I will make this change along with my next PR

PR #174, opened 2023-09-19
renewing sysbench arm support

ericavella:
the arm64/table count issue is fixed! so removing exception against setting num of tables in arm64
also expanding profile scenarios
and to top it all off adding a state property to track if the executables were initialized.


src/VirtualClient/VirtualClient.Main/profiles/PERF-MYSQL-SYSBENCH-OLTP.json
```diff
@@ -27,12 +27,124 @@
...
                 "DatabaseScenario": "$.Parameters.DatabaseScenario",
                 "Role": "Client",
-                "Duration": "00:30:00",
+                "Duration": "02:00:00",
                 "RecordCount": 10000,
                 "Workload": "oltp_read_write",
                 "PackageName": "sysbench",
                 "SkipInitialize": "$.Parameters.SkipInitialize"
             }
+        },
```
yangpanMS, line 36: Thank you for adding this. Does all these work with all-configurations? in-memory and mix etc?

src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPExecutor.cs
```diff
@@ -140,24 +140,37 @@ protected void InitializeApiClients()
...
-                await this.SystemManager.MakeFileExecutableAsync(
-                    this.Combine(scriptsDirectory, "in-memory.sh"),
-                    this.Platform,
-                    cancellationToken);
+                // install sysbench using repo scripts
+                if (this.Platform == PlatformID.Unix && this.DatabaseScenario != SysbenchOLTPScenario.Default)
+                {
+                    string scriptsDirectory = this.PlatformSpecifics.GetScriptPath("sysbencholtp");
+
+                    await this.SystemManager.MakeFileExecutableAsync(
```
yangpanMS, line 155: There is one single method to make all files under one dir executable if you want

src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPExecutor.cs
```diff
@@ -267,6 +280,19 @@ public bool SysbenchInitialized
                 }
             }
 
+            public bool ExecutablesInitialized
```
yangpanMS, line 283: Let's not expose this in parameters. Saving in state is ok, but user shouldn't need to know it

src/VirtualClient/VirtualClient.Main/profiles/PERF-MYSQL-SYSBENCH-OLTP.json
ericavella: yes it does * now * !

src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPExecutor.cs
ericavella: gotcha -- does that mean making this private?

yangpanMS: no I mean making it a property in the MySQL state, so user don't need to know whether they need to pass in this parameter

yangpanMS (approved):

PR #187, opened 2023-10-11
Added support for uploading specviewperf logs to blob storage

LiYueqian-James:
Was able to get log files uploaded using the below vc args:
--profile=PERF-GPU-SPECVIEW-AMD.json  --profile=MONITORS-NONE.json --iterations=1 --packages={SAS} --eventHub={SAS} --cs={SAS} --debug


src/VirtualClient/VirtualClient.Actions.UnitTests/SPEC/SpecViewExecutorTests.cs
```diff
@@ -123,6 +123,11 @@ private async Task SpecViewExecutorExecutesWorkloadAsExpected(PlatformID platfor
                         return this.mockFixture.Process;
                     };
 
+                    // Remove any mock blob managers so that we do not evaluate the code paths that
+                    // upload log files by default.
+                    this.mockFixture.Dependencies.RemoveAll<IEnumerable<IBlobManager>>();
```
yangpanMS, line 128: The tabbing seems odd

yangpanMS (approved):
If you have time, you could add test for file upload, but I won't block the PR on it.

src/VirtualClient/VirtualClient.Actions.UnitTests/SPEC/SpecViewExecutorTests.cs
LiYueqian-James: I left this line inside `if (platform == PlatformID.Win32NT)` as we only suporrt win-x64 for now.

yangpanMS (approved):

PR #196, opened 2023-10-26
Memtier results aggregation

nmalkapuram:
Aggregate metrics for memtier client for both redis and memcached server process.


src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -393,6 +393,8 @@ private async Task ConfigureMakeFileAsync(EventContext telemetryContext, Cancell
                 await this.fileSystem.File.ReplaceInFileAsync(
                         makeFilePath, @"MPlib *=", $"MPlib =  /usr/lib/{architecture}-linux-gnu/openmpi/lib/libmpi.so", cancellationToken);
 
+                // /usr/lib/x86_64-linux-gnu/openmpi
```
yangpanMS, line 396: Does this comment match the command?

src/VirtualClient/VirtualClient.Actions/Memtier/MemtierBenchmarkClientExecutor.cs
```diff
@@ -346,6 +368,85 @@ private void CaptureMetrics(string results, string commandArguments, DateTime st
             }
         }
 
+        private void CaptureAggregateMetrics(string commandArguments, DateTime startTime, DateTime endTime, EventContext telemetryContext, CancellationToken cancellationToken)
```
yangpanMS, line 371: Can you do the aggregation in the parser class? This allows us to properly unit test it

src/VirtualClient/VirtualClient.Actions/Memtier/MemtierBenchmarkClientExecutor.cs
```diff
@@ -6,20 +6,23 @@ namespace VirtualClient.Actions
...
     using System.Diagnostics;
+    using System.Linq;
     using System.Net;
     using System.Runtime.InteropServices;
     using System.Text.RegularExpressions;
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Extensions.DependencyInjection;
     using Microsoft.Extensions.Logging;
+    using Microsoft.Identity.Client;
```
yangpanMS, line 17: Not sure if you need this

yangpanMS (changes requested):
For metric changes could you please add changes for both metric documents and also the unit test?

src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
nmalkapuram: It does with one of the commands that is used for replacing in makeFile. But it is not needed in this PR. I will remove it.

yangpanMS:
Also please rename the title

src/VirtualClient/VirtualClient.Actions/Memtier/MemtierBenchmarkClientExecutor.cs
nmalkapuram: Here we aren't using results/metrics from single object of Parser and aggregating the metrics. We are aggregating the metrics across results of different parser objects.

src/VirtualClient/VirtualClient.Actions/Memtier/MemtierBenchmarkClientExecutor.cs
nmalkapuram: Removing it. Thank you

src/VirtualClient/VirtualClient.Actions/Memtier/MemtierBenchmarkClientExecutor.cs
yangpanMS: You can still do it with Parser, the constructor is free, you can change that to params[] string or oput the Aggregate method on Parser, or something else

src/VirtualClient/VirtualClient.Actions.UnitTests/Examples/Memtier/RedisResults_3.txt
```diff
@@ -0,0 +1,392 @@
...
+4         Threads
+1         Connections per thread
+30        Seconds
+
+
+ALL STATS
+============================================================================================================================================================
+Type         Ops/sec     Hits/sec   Misses/sec    Avg. Latency     p50 Latency     p90 Latency     p95 Latency     p99 Latency   p99.9 Latency       KB/sec
+------------------------------------------------------------------------------------------------------------------------------------------------------------
+Sets        35598.74          ---          ---         0.34304         0.33500         0.47900         0.55900         0.83900         1.54300      2742.53
```
yangpanMS, line 11: I don't see an example where you need aggregation, could you add that?

src/VirtualClient/VirtualClient.Actions.UnitTests/Memtier/MemtierMetricsParserTests.cs
```diff
@@ -15,16 +16,148 @@ namespace VirtualClient.Actions
...
+            MetricAssert.Exists(metrics, "GET_Latency-P95_Stdev", 0, MetricUnit.Milliseconds);
+            MetricAssert.Exists(metrics, "GET_Latency-P95_P80", 4.41500, MetricUnit.Milliseconds);
+
+            MetricAssert.Exists(metrics, "GET_Latency-P99_Avg", 7.42300, MetricUnit.Milliseconds);
+            MetricAssert.Exists(metrics, "GET_Latency-P99_Min", 7.42300, MetricUnit.Milliseconds);
+            MetricAssert.Exists(metrics, "GET_Latency-P99_Max", 7.42300, MetricUnit.Milliseconds);
+            MetricAssert.Exists(metrics, "GET_Latency-P99_Stdev", 0, MetricUnit.Milliseconds);
+            MetricAssert.Exists(metrics, "GET_Latency-P99_P80", 7.42300, MetricUnit.Milliseconds);
+
+            MetricAssert.Exists(metrics, "GET_Latency-P99.9_Avg", 29.31100, MetricUnit.Milliseconds);
```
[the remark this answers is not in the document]
nmalkapuram: We technically don't need aggregation if we have single set of results. We will have 1 set of results only when we run on 1vcpu machine. I believe we usually run on way larger machines with more than 1 vcpus. Yes in that case it's multiplying count of metrics by almost 5 not exactly 5.  

For every set of results we get 26 different per process metrics. As we run on higher vcpu machine, the count of output metrics increase linearly ( 26*count of vcpus ). Now with the aggregate it's 138 metrics irrespective of number of vcpus. 

Do you want me to handle that case in which we get only one set of result so that we don't aggregate in that scenario?

src/VirtualClient/VirtualClient.Actions.UnitTests/Examples/Memtier/RedisResults_3.txt
nmalkapuram: Sure. Aggregation is across different output files generated from each client communicating to different Redis server processes running on different ports. Total number of output files generated will be equal to VCPUs*clientInstanceCount. We are aggregating here to know how all the Redis processes are running on a system as a whole. Metrics from individual Redis processes have become a little less important than the overall aggregate metrics. Anyway, I will change unit tests to accept multiple set of results for aggregation. Thanks Yang.

yangpanMS: Thanks I just want to see a set where it is different

yangpanMS (approved):

PR #200, opened 2023-11-02
Users/prachisingla/fix disk path issue

psingla1210:


src/VirtualClient/VirtualClient.Contracts/DiskExtensions.cs
```diff
@@ -200,7 +200,7 @@ public static bool PathEquals(this Disk disk, string path)
 
             if (disk.Volumes?.Any() == true)
             {
-                volumeMatchDevicePath = disk.Volumes.Any(v => string.Equals(v.DevicePath.TrimEnd('/', '\\', ':'), path, StringComparison.OrdinalIgnoreCase));
+                volumeMatchDevicePath = disk.Volumes.Any(v => string.Equals(v.DevicePath?.TrimEnd('/', '\\', ':'), path, StringComparison.OrdinalIgnoreCase));
                 volumeMatchAccessPath = disk.Volumes.Any(v => v.AccessPaths.Any(ap => string.Equals(ap.TrimEnd('/', '\\', ':'), path, StringComparison.OrdinalIgnoreCase)));
```
yangpanMS, line 204: Would we need to update line 204 too? I didn't face issue with line 203. Did you hit issues?

psingla1210: DevicePath is allowed to be null, so we need to do this. AccessPaths can not be null, so that's why did not add there.

yangpanMS (approved):

PR #206, opened 2023-11-09
removing existing compiler alternatives prior to adding them

ericavella:
During Cedar Crest validation, we ran into a bug that would not allow us to get past the CompilerInstallation dependency on the linux-arm64 system.
image

By removing all alternatives before adding new ones, this error is avoided.
This change was also tested on fresh linux-x64 and -arm64 systems to assure the change is non-breaking, as well as the original CC machine.


src/VirtualClient/VirtualClient.Dependencies/CompilerInstallation.cs
```diff
@@ -237,18 +236,44 @@ private async Task InstallGccAsync(string gccVersion, EventContext telemetryCont
...
+                try
+                {
+                    await this.ExecuteCommandAsync("update-alternatives", $"--remove-all {compiler}", Environment.CurrentDirectory, telemetryContext, cancellationToken);
+                }
+                catch 
+                {
+                    // the message is:
+                    //      "error: no alternatives for g++"
+                    // so we can continue as normal; non-breaking
+                    continue;
```
yangpanMS, line 264: Technically you don't need the continue, but I don't mind either way

src/VirtualClient/VirtualClient.Dependencies/CompilerInstallation.cs
```diff
@@ -237,18 +236,44 @@ private async Task InstallGccAsync(string gccVersion, EventContext telemetryCont
 
         private async Task SetGccPriorityAsync(string gccVersion, EventContext telemetryContext, CancellationToken cancellationToken)
         {
+            string[] compilers =
+            {
+                "g++",
```
yangpanMS, line 241: add gcc and cc

src/VirtualClient/VirtualClient.Dependencies.UnitTests/CompilerInstallationTests.cs
```diff
@@ -69,13 +69,19 @@ public async Task CompilerInstallationRunsTheExpectedWorkloadCommandInLinuxForGc
                 "sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y",
                 "sudo apt update",
                 "sudo apt install build-essential gcc-123 g++-123 gfortran-123 -y --quiet",
+                "sudo update-alternatives --remove-all g++",
```
yangpanMS, line 72: do you need to change this accordingly?

yangpanMS (approved):

PR #220, opened 2023-11-22
remove compilers fix

ericavella:
separating out into stages,
not removing alternatives for cc


src/VirtualClient/VirtualClient.Dependencies/CompilerInstallation.cs
```diff
@@ -235,6 +236,37 @@ private async Task InstallGccAsync(string gccVersion, EventContext telemetryCont
...
 
+        private async Task RemoveAlternativesAsync(EventContext telemetryContext, CancellationToken cancellationToken)
+        {
+            string[] compilers =
+            {
+                "gcc",
+                "g++",
+                "gcov",
+                "gcc-ar",
+                "gcc-ranlib",
```
yangpanMS, line 247: could you research if the gcc installation also adds link for gcov gcc-ar, gcc-ranlib? if so let's not remove these as well

ericavella: afaik gcc installation doesn't add any links
but the test vms had gcov as a master alternative set already

yangpanMS: use brand new ubuntu vm and let's not use cc test vm to be honest I think we only need to remove gcc and gfortran, it would be great if you could test it out

yangpanMS (approved):

PR #222, opened 2023-11-28
Resolving graph500 compilation issue

nmalkapuram:


src/VirtualClient/VirtualClient.Main/profiles/PERF-GRAPH500.json
```diff
@@ -5,18 +5,20 @@
...
         "SupportedOperatingSystems": "CBL-Mariner,CentOS,Debian,RedHat,Suse,Ubuntu"
     },
-    "Parameters": {
-        "Scale": "10",
-        "EdgeFactor": "16"
-    },
+  "Parameters": {
+    "Scale": "10",
+    "EdgeFactor": "16",
+    "CommandArguments": "LDFLAGS=\"-lpthread -fcommon\""
```
yangpanMS, line 11: Just to make sure, is this for a specific compiler version or architecture? Also these could be compiler flags instead of CommandArguments, where is usually intended for workloads.

nmalkapuram: The reason I found compilation issue happens is because of how global variables are declared in the workload. 
A common mistake in C is omitting **extern** when declaring a global variable in a header file. If the header is included by several files it results in multiple definitions of the same variable. In previous GCC versions this error is ignored. GCC 10 defaults to **-fno-common**, which means a linker error will now be reported. GCC version above 10 will have this issue and can be removed by using -fcommon flag. -lpthread flag is irrespective of the compiler version. But even for the GCC versions less than 10 having -fcommon flag doesn't have any effect. 
In graph500 we aren't explicitly controlling gcc versions, we actually download mpich which by defaults downloads mpicc compiler which is the implementation around gcc. The version of mpich are like 3.4, 4. Version 4.0 installs 11 version of mpicc or gcc

yangpanMS (approved):

PR #227, opened 2023-11-30
Users/nmalkapuram/memtier parser bug

nmalkapuram:
Parser update for the general output from memtier which was effecting aggregate metrics.
commandline update for aggregate metrics.
VC version update


src/VirtualClient/VirtualClient.Actions/Memtier/MemtierMetricsParser.cs
```diff
@@ -29,7 +29,7 @@ public class MemtierMetricsParser : MetricsParser
         // Totals      48271.29     43444.12         0.00         2.62213         2.75100         3.90300         4.41500         7.42300        29.31100      4053.46
         private static readonly Regex GetsExpression = new Regex(@"(?<=Gets).*(?=\n)", RegexOptions.IgnoreCase);
         private static readonly Regex SetsExpression = new Regex(@"(?<=Sets).*(?=\n)", RegexOptions.IgnoreCase);
-        private static readonly Regex TotalsExpression = new Regex(@"(?<=Totals).*(?=\n)", RegexOptions.IgnoreCase);
```
yangpanMS, line 32: Up to you but if you have seen something with white space, you can add {\n}+ if needed

nmalkapuram: I believe that shouldn't be needed. After totals line we would either have a new line or nothing. Even if we see whitespaces current parser is able to handle this scenario. * at the end of Totals looks for any character except for new line.

yangpanMS (approved):

PR #229, opened 2023-12-01
Fix for double quotes appearing in csv log

imadityaa:
Virtual Client logs the produced metrics locally in csv format. If any of the fields appearing in the csv has quotes ("), it can potentially break the csv. 
As a standard way of handling writing to csv, double quote appearing in the field should be escaped by preceding it with another double quote.
Reference: https://stackoverflow.com/questions/17808511/how-to-properly-escape-a-double-quote-in-csv

Modifying one the existing Unit Tests as well to have this unit tested.


src/VirtualClient/VirtualClient.Core/Logging/MetricsCsvFileLogger.cs
```diff
@@ -169,7 +169,7 @@ internal static string CreateMessage(EventContext context)
         {
             StringBuilder messageBuilder = new StringBuilder();
             messageBuilder.Append(Environment.NewLine);
-            messageBuilder.AppendJoin(',', MetricsCsvFileLogger.CsvFields.Select(field => $"\"{field.GetFieldValue(context)}\""));
+            messageBuilder.AppendJoin(',', MetricsCsvFileLogger.CsvFields.Select(field => $"\"{field.GetFieldValue(context)?.Replace("\"", "\"\"")}\""));
```
yangpanMS, line 172: Thank you, which workload have we seen failures here?

yangpanMS (approved):
Approved, but is there any other char that we need to escape?

PR #237, opened 2023-12-07
sysbench threads ceiling

ericavella:
Sysbench thread count has a ceiling -- protecting against failures for VMs with high CPU count by setting a max threads value.


src/VirtualClient/VirtualClient.Actions/SysbenchOLTP/SysbenchOLTPClientExecutor.cs
```diff
@@ -157,6 +157,8 @@ public int Threads
                 {
                     numThreads = threads.ToInt32(CultureInfo.InvariantCulture);
                 }
+
```
yangpanMS, line 160: nit: Math.Max is more intuitive

yangpanMS (approved):

yangpanMS:
Also please help make sure we have unit test covering this usage. Also add one-liner in our document

yangpanMS (approved):

PR #239, opened 2023-12-26
Fixing bugs in lmbench and hplinpack

nmalkapuram:


src/VirtualClient/VirtualClient.Main/profiles/PERF-MEM-LMBENCH.json
```diff
@@ -1,42 +1,55 @@
...
+  "MinimumExecutionInterval": "00:10:00",
+  "Metadata": {
+    "RecommendedMinimumExecutionTime": "(4-cores)=04:00:00,(16-cores)=10:00:00,(64-cores)=16:00:00",
+    "SupportedPlatforms": "linux-x64,linux-arm64",
+    "SupportedOperatingSystems": "CBL-Mariner,CentOS,Debian,RedHat,Suse,Ubuntu"
+  },
+  "Parameters": {
+    "CompilerName": "gcc",
+    "CompilerVersion": "10",
+    "CompilerFlags": "CPPFLAGS=\"-I /usr/include/tirpc\"",
```
yangpanMS, line 12: Which systems need these flags?

nmalkapuram: ubuntu 22 both x64 and arm64 platforms require these flags.

yangpanMS: which systems are you testing on? I remember we have ubuntu22 data on lmbench

yangpanMS, line 12: Could you document the implication of the flags and what are the perf impact on the non-ubuntu22 systems?

nmalkapuram: I was testing on azure VMs ubuntu22 arm machines. Also facing same issue on cedarcrest machines.

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
```diff
@@ -43,7 +43,7 @@
       "Parameters": {
         "Scenario": "InstallRequiredLinuxPackagesForHPL",
         "Repositories-Apt": "ppa:ubuntu-toolchain-r/test",
-        "Packages": "build-essential,linux-generic,libmpich-dev,libopenmpi-dev,libatlas-base-dev"
+        "Packages": "build-essential,libmpich-dev,libopenmpi-dev,libatlas-base-dev"
```
yangpanMS, line 46: That's a lot of packages, does this only work in ubuntu or apt store? How about other distro?

src/VirtualClient/VirtualClient.Main/profiles/PERF-MEM-LMBENCH.json
yangpanMS, line 12: Also have you tried if you need these in other compiler versions?

nmalkapuram: Yes, same issue is across different versions of compiler.

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
nmalkapuram: Adding packages for other distros too.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MEM-LMBENCH.json
nmalkapuram: In some systems after installing the required packages few links for the compiler are by default set as part of installation process. But on ubuntu 22 systems, these links to the compiler are not set at the time of installation and needed to be explicitly passed as flags to the compiler to set up the links. 
Compilation itself throws error displaying required files are not found. 
fatal error: rpc/rpc.h: No such file or directory. These files/libraries are used as part of lmbench compilation.

PR #244, opened 2024-01-08
Sockperf reset instruction from client

nmalkapuram:
SockPerf Server doesn't receive reset instruction from client at the end of client's execution and client keeps on waiting for the reset server (stopping server and state deletion). Adding the reset instruction from client to server.


nmalkapuram:
Created a new PR adding additional fixes

PR #252, opened 2024-01-15
Users/nmalkapuram/user name detection logic

nmalkapuram:


src/VirtualClient/VirtualClient.Core/SystemManagement.cs
```diff
@@ -125,6 +125,22 @@ public async Task<CpuInfo> GetCpuInfoAsync(CancellationToken cancellationToken)
             return info;
         }
 
+        /// <summary>
+        /// Get the logged In Username i.e, username of the user who invoked a command with elevated privileges using the "sudo" command in Unix operating system.
+        /// </summary>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public string GetLoggedInUserName(CancellationToken cancellationToken)
```
deep1712, line 133: Why do we need cancellation token?

nmalkapuram: Removing it.

src/VirtualClient/VirtualClient.Actions/Memcached/MemcachedServerExecutor.cs
```diff
@@ -5,6 +5,7 @@ namespace VirtualClient.Actions
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics;
```
psingla1210, line 8: NIT - this file can be removed from change, right?

psingla1210 (approved):

src/VirtualClient/VirtualClient.Core/SystemManagement.cs
```diff
@@ -125,6 +125,20 @@ public async Task<CpuInfo> GetCpuInfoAsync(CancellationToken cancellationToken)
...
 
+        /// <summary>
+        /// Get the logged In Username i.e, username of the user who invoked a command with elevated privileges using the "sudo" command in Unix operating system.
+        /// </summary>
+        public string GetLoggedInUserName()
+        {
+            string loggedInUserName = Environment.GetEnvironmentVariable("SUDO_USER");
+            if (string.IsNullOrEmpty(loggedInUserName))
+            {
+                loggedInUserName = Environment.UserName;
```
brdeyo, line 136: Double-check for me on this but it was my understanding that Yang had validated that "Environment.UserName" by itself works. I remember there might be a minor distinction when we are launching VC on a VM from the Juno Guest agent which runs as a service/daemon. 

Can you clarify the distinction here?

nmalkapuram: When we manually run VC on VM and run using sudo, Environment.UserName returns "root" on linux. We don't get the actual username who is logged in but we get "root". There are certain cases where we need to use the actual username. e.g: azureuser.

yangpanMS: When I tried I remember Environemnt.Username worked. Maybe this is distro-version specific

Can we try this? Read Environment.Username first, if it's root, then we read the viarable, and return the one that doesn't say root

yangpanMS: Let's take this offline. I believe this is only in Juno, where root is the true invoking user.

yangpanMS:
As confirmed, setting to draft. Please convert back once ready

src/VirtualClient/VirtualClient.Core/SystemManagement.cs
```diff
@@ -125,6 +125,28 @@ public async Task<CpuInfo> GetCpuInfoAsync(CancellationToken cancellationToken)
...
+            string loggedInUserName = Environment.UserName;
+            if (string.Equals(loggedInUserName, "root"))
+            {
+                loggedInUserName = Environment.GetEnvironmentVariable("SUDO_USER");
+                if (string.Equals(loggedInUserName, "root") || string.IsNullOrEmpty(loggedInUserName))
+                {
+                    loggedInUserName = Environment.GetEnvironmentVariable("VC_SUDO_USER");
+                    if (string.IsNullOrEmpty(loggedInUserName))
+                    {
+                        throw new EnvironmentSetupException($"VC_SUDO_USER Environment variable is expected to be set.", ErrorReason.EnvironmentIsInsufficent);
```
yangpanMS, line 143: Let's give a more comprehensive error including all the 3 methods we used.

src/VirtualClient/VirtualClient.Dependencies/Packaging/WgetPackageInstallation.cs
```diff
@@ -122,6 +122,12 @@ protected override async Task ExecuteAsync(EventContext telemetryContext, Cancel
                         string installationPath = downloadedPackagePath;
                         string packagesDirectory = this.GetPackagePath();
 
+                        // Create packages directory if not present.
+                        if (!this.fileSystem.Directory.Exists(packagesDirectory))
+                        {
+                            this.fileSystem.Directory.CreateDirectory(packagesDirectory).Create();
```
yangpanMS, line 128: Do we need the additional Create()at the end? Or the createDirectory will just do it?

yangpanMS (approved):

yangpanMS (approved):

src/VirtualClient/VirtualClient.Dependencies/Packaging/WgetPackageInstallation.cs
nmalkapuram: yes. Just CreateDirectory() is sufficient.  Changing it.

yangpanMS (approved):

PR #255, opened 2024-01-26
Backlogs changes

LiYueqian-James:
1. Added support for specview individual test name.
2. Fixed a bug in nvidiaDriverInstalltion
3. Renamed blenderbenchmark pkg name for better consistency
4. Set default GpuModel to empty to prevent installing wrong driver accidentally.


src/VirtualClient/VirtualClient.Main/profiles/PERF-BLENDER-AMD.json
```diff
@@ -16,7 +16,7 @@
       "Type": "BlenderBenchmarkExecutor",
       "Parameters": {
         "Scenario": "BlenderCPUAndGPUPerformance",
-        "PackageName": "blenderbenchmarkcli",
+        "PackageName": "blenderbenchmark",
```
yangpanMS, line 19: You also need to update line 51

src/VirtualClient/VirtualClient.Dependencies/CUDAAndNvidiaGPUDriverInstallation.cs
```diff
@@ -184,7 +184,7 @@ await this.stateManager.SaveStateAsync(nameof(CudaAndNvidiaGPUDriverInstallation
                     await this.CudaAndNvidiaGPUDriverInstallationOnWindowsAsync(telemetryContext, cancellationToken)
                                .ConfigureAwait(false);
 
-                    await this.stateManager.SaveStateAsync(nameof(this.CudaAndNvidiaGPUDriverInstallationOnWindowsAsync), new State(), cancellationToken)
+                    await this.stateManager.SaveStateAsync(nameof(CudaAndNvidiaGPUDriverInstallation), new State(), cancellationToken)
```
yangpanMS, line 187: Did this break on you? This should work if you are only working on Windows and not switching OS

LiYueqian-James: on windows, if I run multiple iterations of installGPUDriver, the first iteration will install the driver successfully, but the following iterations will not skip the installation as expected. I saw that on [line 147](https://github.com/microsoft/VirtualClient/blob/b91e115e177d733b8911a4fc0175b420e08be7cc/src/VirtualClient/VirtualClient.Dependencies/CUDAAndNvidiaGPUDriverInstallation.cs#L147), the executor is getting the state using the name CudaAndNvidiaGPUDriverInstallation, not CudaAndNvidiaGPUDriverInstallationOnWindowsAsync. After changing the variable name, VC skips execution after the first run as expected.

yangpanMS (approved):

yangpanMS (approved):

PR #256, opened 2024-01-26
Updated JDK package installation into one dependency

YeshwanthReddySangam:
Combined JavaDevelopmentKitInstallation and DependencyPackageInstallation into one single dependency

### DIFF
--- a/src/VirtualClient/VirtualClient.Dependencies/JDKPackageDependencyInstallation.cs
+++ b/src/VirtualClient/VirtualClient.Dependencies/JDKPackageDependencyInstallation.cs
@@ -12,22 +12,23 @@ namespace VirtualClient.Dependencies
     using VirtualClient.Common.Extensions;
     using VirtualClient.Common.Telemetry;
     using VirtualClient.Contracts;
+    using VirtualClient.Dependencies.Packaging;
 
     /// <summary>
     /// Provides functionality for installing the JDK on the system.
     /// </summary>
-    public class JavaDevelopmentKitInstallation : VirtualClientComponent
+    public class JDKPackageDependencyInstallation : DependencyPackageInstallation
     {
         private ISystemManagement systemManager;
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


YeshwanthReddySangam:

website/docs/dependencies/0050-install-java-jdk.md
```diff
@@ -32,31 +32,27 @@ The following section describes the parameters used by the individual component
 
 | **Parameter** | **Required** | **Description**                                                                                                 |
 |---------------|--------------|-----------------------------------------------------------------------------------------------------------------|
```
yangpanMS, line 34: Thanks for updating the document. Please add 1/2 sentences about how this derives from the DependencyPackageInstallation

yangpanMS (approved):
We don't really have this scenario yet but this now forbid users from using their own packaged JDK. It has to be downloaded. Can we find a path where if the blob is not defined, the JdkInstallation proceeds to find the package via package name?

src/VirtualClient/VirtualClient.Dependencies/JDKPackageDependencyInstallation.cs
```diff
@@ -45,6 +46,8 @@ public JavaDevelopmentKitInstallation(IServiceCollection dependencies, IDictiona
         /// </summary>
         protected override async Task ExecuteAsync(EventContext telemetryContext, CancellationToken cancellationToken)
         {
+            // Install JDK package
+            await base.ExecuteAsync(telemetryContext, cancellationToken);
```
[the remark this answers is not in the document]
YeshwanthReddySangam: Updated documentation for transition between versions

website/docs/dependencies/0050-install-java-jdk.md
YeshwanthReddySangam: Updated documentation along with installation steps

yangpanMS (approved):
Great work, thanks!

PR #264, opened 2024-02-08
Adding support for H100 for MLPerf

psingla1210:


src/VirtualClient/VirtualClient.Main/profiles/PERF-GPU-MLPERF-NVIDIA-H100.json
```diff
@@ -0,0 +1,93 @@
...
+    "Actions": [
+        {
+            "Type": "MLPerfExecutor",
+            "Parameters": {
+                "Scenario": "inferencemodels-bert",
+                "Model": "bert",
+                "Username": "$.Parameters.Username",
+                "DiskFilter": "$.Parameters.DiskFilter",
+                "RequireCustomSystemSupport": "$.Parameters.RequireCustomSystemSupport",
+                "Benchmarks": "$.Parameters.Benchmarks"
```
yangpanMS, line 29: I don't understand here why you have separate scenario names but they are running the same benchmark

src/VirtualClient/VirtualClient.Main/profiles/PERF-GPU-MLPERF-NVIDIA-H100.json
```diff
@@ -0,0 +1,93 @@
+{
+    "Description": "MLPerf GPU Performance Workload",
+    "Metadata": {
+        "RecommendedMinimumExecutionTime": "10:00:00",
+        "SupportedPlatforms": "linux-x64",
+        "SupportedOperatingSystems": "Ubuntu",
+        "SpecialRequirements": "This is an Nvidia GPU-specialized workload. It depends upon the system having an Nvidia GPU card/chip.",
+        "GPU": "H100"
```
yangpanMS, line 8: Is this referenced? Why do we need a separate profile for H100?

website/docs/workloads/mlperf/mlperf-profiles.md
```diff
@@ -38,6 +38,7 @@ Runs the MLPerf benchmark workload to test GPU performance.
   | LinuxDriverVersion     | Optional. Version of GPU driver that needs to be installed. | 510  |
   | LinuxLocalRunFile     | Optional. Link to download specified CUDA and GPU driver versions. | https://developer.download.nvidia.com/compute/cuda/12.0.0/local_installers/cuda_12.0.0_525.60.13_linux.run  |
   | RequireCustomSystemSupport | Optional. This enables additional A100_PCIe_80GBx4 system support that was not supported by github repo of MLPerf. Ones that are supported by github repo of MLPerf are still supported. | true |
```
yangpanMS, line 40: I would like to see if this could be removed now. Is MLPerf supporting A100x4 now?

src/VirtualClient/VirtualClient.Main/profiles/PERF-GPU-MLPERF-NVIDIA-H100.json
psingla1210: Model parameter determines which benchmark will be run, this benchmarks parameter is the list of all benchmarks that will be run as part of complete profile, reason being we prepare the setup by downloading all the required datasets and then prebuild for those required benchmarks and then each model is run

src/VirtualClient/VirtualClient.Main/profiles/PERF-GPU-MLPERF-NVIDIA-H100.json
psingla1210: List of benchmarks supported by MLPerf for A series and then H series is different. This brings a good point that i should rename to Hseries instead of H100. To keep things clear for user, I have created a separate profile. I know that this is not a sustainable approach but in my opinion we can keep MLPerf in our exceptional benchmarks bucket that needs attention with new GPU HW coming in. As of now, quite a few people are interested and will be interested to run MLPerf for H100 and then probably B200 and then later do comparisons among different, so to keep things simple to run, I have created separate profiles for them.

website/docs/workloads/mlperf/mlperf-profiles.md
psingla1210: No, they are not.

src/VirtualClient/VirtualClient.Main/profiles/PERF-GPU-MLPERF-NVIDIA-H100.json
yangpanMS: Can we prepare the dataset only for the current defined workload? Instead of everything upfront?

yangpanMS:
Abandoning as discussed

PR #271, opened 2024-02-20
sysbench components + scripts

ericavella:
separating sysbench into multiple components
delocalizing database population & distribution
utilizing package store
leaning on python scripts to perform command work


src/VirtualClient/VirtualClient.Main/profiles/PERF-MYSQL-OLTP-SYSBENCH.json
```diff
@@ -8,135 +8,169 @@
...
+        "NumTables": "10",
+        "DatabaseScenario": "Default",
+        "Duration": "00:05:00",
+        "SQLServerPackageName": "mysql-server",
+        "WorkloadGeneratorPackageName": "sysbench"
     },
     "Actions": [
         {
-            "Type": "SysbenchOLTPServerExecutor",
+            "Type": "SysbenchConfiguration",
```
brdeyo, line 19: I would still expect the database initial creation, distribution across the disks and then the full population to be in the Dependencies. I don't think we need to support changing the database size on each iteration of the profile. If the user needs to support more than 1 database size on the same system, I recommend that we support this in the following way in the near term (vs. trying to do this mid-stream in the Actions):

How to test against different database sizes:
User will simply perform a VC reset (i.e. VirtualClient.exe clean, VirtualClient.exe reset) in between the runs. A reset/clean removes all packages that are not "built-in" packages, deletes the state files in the /state directory and deletes the log files in the /logs directory. The intention is to set VC back to the state it was in at the very beginning before being run.

We should support this behavior of course. So for example, if the user starts with a database that has 10 tables with 10,000,000 records each and runs. The user should then be able to do a VC clean and execute a second scenario where there are 40 tables with 40,000,000 records each. We should be able to handle the following in support of this:

1) Repeat installations of MySQL server software.
2) Repeat updates to the MySQL variables in the config file for the server.
3) Resetting the database. Given that Sysbench simply appends to any existing database tables, we either need to drop the DB and rebuild or have a way to check to see how many tables and records there are. The latter sounds complicated, so I vote for the former...keep it simple.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MYSQL-OLTP-SYSBENCH.json
```diff
@@ -8,135 +8,169 @@
     },
     "Parameters": {
         "DatabaseName": "sbtest",
-        "DatabaseScenario": "Balanced",
-        "Duration": "00:05:00"
+        "NumTables": "10",
```
brdeyo, line 11: Couple of things:
1) Naming -> TableCount

2) The user will need to have the ability to manage the number of records in each of the tables as well. I see the conundrum here between supporting 3 scenarios: 1) balanced, 2) in-memory or 3) user-specified # of tables and records. So, let's just make a decision now. For OUR profile, we will support the DatabaseScenario (Balanced or InMemory) out of the box. However, we will still support the ability of the user to supply a TableCount and RecordCount parameter in the SysbenchConfiguration for populating the database. We just won't expose it here in our profile. The ABC team can create their own profiles if they want to have a database of some specific size. This profile is NOT for their purposes.

Let's sync on this when you have a sec so we are on both in agreement with our "pattern"

src/VirtualClient/VirtualClient.Actions/Sysbench/SysbenchExecutor.cs
```diff
@@ -48,7 +49,53 @@ public string DatabaseScenario
...
             {
-                return this.Parameters.GetValue<string>(nameof(this.DatabaseScenario), SysbenchOLTPScenario.Default);
+                return this.Parameters.GetValue<string>(nameof(this.DatabaseScenario), SysbenchScenario.Default);
+            }
+        }
+
+        /// <summary>
+        /// Number of records per table.
+        /// </summary>
+        public int RecordCount
```
brdeyo, line 59: Code Implementation Recommendation: Avoid significant logic in C# property getters and setters.

Move the logic here for defining the RecordCount to the InitializeAsync() method to be consistent with the relative convention in other VC components.

src/VirtualClient/VirtualClient.Actions/Sysbench/SysbenchConfiguration.cs
```diff
@@ -0,0 +1,147 @@
...
+            get
+            {
+                return this.Parameters.GetValue<string>(nameof(SysbenchConfiguration.DiskFilter), string.Empty);
+            }
+        }
+
+        /// <summary>
+        /// The workload option passed to Sysbench.
+        /// </summary>
+        public int TableCount
```
brdeyo, line 52: Code Implementation Recommendation: Avoid significant logic in C# property getters and setters.

Move the logic here for defining the TableCount to the InitializeAsync() method to be consistent with the relative convention in other VC components.

src/VirtualClient/VirtualClient.Actions/Sysbench/SysbenchClientExecutor.cs
```diff
@@ -0,0 +1,281 @@
...
+            get
+            {
+                return this.Parameters.GetValue<string>(nameof(SysbenchClientExecutor.Workload));
+            }
+        }
+
+        /// <summary>
+        /// The workload option passed to Sysbench.
+        /// </summary>
+        public int TableCount
```
brdeyo, line 74: Code Implementation Recommendation: Avoid significant logic in C# property getters and setters.

Move the logic here for defining the TableCount to the InitializeAsync() method to be consistent with the relative convention in other VC components.

In fact, given we are using similar/same logic in multiple components (e.g. SysbenchClientExecutor, SysbenchConfiguration, SysbenchExecutor), I recommend creating a static properties and methods on the SysbenchConfiguration class to handle this logic.

e.g.
public const int DefaultTableCount = 10;

public static int GetTableCount(ISystemManagement systemManagement, string databaseScenario);
public static int GetRecordCount(ISystemManagement systemManagement, string databaseScenario);

...then use these statics in each of the individual components (i.e. in the InitializeAsync method) to set the TableCount and RecordCount properties to values (if they are not already set).

brdeyo (changes requested):

brdeyo (approved):

PR #280, opened 2024-03-14
Users/prachisingla/adding more metrics for nvidia smi

psingla1210:


src/VirtualClient/VirtualClient.Monitors/NvidiaSmiQueryGpuParser.cs
```diff
@@ -43,12 +51,38 @@ public override IList<Metric> Parse()
...
                     { "pcie.link.gen.current", Convert.ToString(row[6]) }
                 };
-                metrics.Add(new Metric("temperature.gpu", Convert.ToDouble(row[7]), unit: "celsuis", metadata: metadata));
-                metrics.Add(new Metric("utilization.gpu [%]", Convert.ToDouble(row[8]), unit: "%", metadata: metadata));
-                metrics.Add(new Metric("utilization.memory [%]", Convert.ToDouble(row[9]), unit: "%", metadata: metadata));
-                metrics.Add(new Metric("memory.total [MiB]", Convert.ToDouble(row[10]), unit: "MiB", metadata: metadata));
-                metrics.Add(new Metric("memory.free [MiB]", Convert.ToDouble(row[11]), unit: "MiB", metadata: metadata));
-                metrics.Add(new Metric("memory.used [MiB]", Convert.ToDouble(row[12]), unit: "MiB", metadata: metadata));
+
+                metrics.Add(new Metric("utilization.gpu", Convert.ToDouble(row[7]), unit: "%", metadata: metadata));
```
yangpanMS, line 55: Is there a way to parse the table properly instead of relying on column count? Just in case some column name switches and we will give wrong data.

yangpanMS (approved):

src/VirtualClient/VirtualClient.Monitors/NvidiaSmiQueryGpuParser.cs
psingla1210: That is a good point. Now parsing using column name. Thank you :)

yangpanMS (approved):

PR #290, opened 2024-03-31
Added Flush All Executor for Redis.

RakeshwarK:


yangpanMS:
Could you also help add document this class, and also update the json profile if needed? That will help me understand why we need this change

RakeshwarK:
Please hold on review, still on discussion about ways to implement flush all for Redis.

PR #296, opened 2024-04-11
JDK path changes

saibulusu:
Environment variable JAVA_HOME does not include 'bin' in path.

### DIFF
--- a/.pipelines/azure-pipelines.yml
+++ b/.pipelines/azure-pipelines.yml
@@ -7,81 +7,84 @@
 trigger:
   branches:
     include:
-      - main
+    - main
   paths:
     exclude:
-      - website
-
+    - website
+    
 pr: none
 
 pool:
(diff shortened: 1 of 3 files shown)


.pipelines/azure-pipelines.yml
```diff
@@ -7,84 +7,81 @@
 trigger:
   branches:
     include:
-    - main
+      - main
```
yangpanMS, line 10: Just to make sure those space weren't intended. Might break the yml

yangpanMS (approved):
Approved but check the yaml and see if you changed anything other than the version

yangpanMS (approved):

.pipelines/azure-pipelines.yml
saibulusu: Pushed a second commit which resolves this issue.

PR #311, opened 2024-05-11
Auto-increment version

saibulusu:
Automatically increment patch number of version within the file VERSION whenever it is not manually updated.


.github/workflows/increment-version.yml
```diff
@@ -0,0 +1,22 @@
+on:
```
yangpanMS, line 1: Please add Action name and also description for the job, to explain what this pipeline does

yangpanMS (approved):

.github/workflows/increment-version.yml
saibulusu: Added a comment with a description of this pipeline. Where should I add the name?

yangpanMS (approved):

PR #341, opened 2024-07-08
Added a new optional parameter ClientsMax to Memtier Client Executor.

RakeshwarK:
This feature allows the user to set Max number of Clients to be used.


src/VirtualClient/VirtualClient.Actions/Memtier/MemtierBenchmarkClientExecutor.cs
```diff
@@ -155,6 +155,17 @@ public string RedisResourcesPackageName
             }
         }
 
+        /// <summary>
+        /// Parameter defines the maximum number of memtier processes.
+        /// </summary>
+        public int ClientsMax
```
yangpanMS, line 161: nit: please rename to MaxClients

yangpanMS (approved):

src/VirtualClient/VirtualClient.Actions/Memtier/MemtierBenchmarkClientExecutor.cs
RakeshwarK: Updated, can you please re-approve?

PR #345, opened 2024-07-15
Added HPLinpack Arm Performance Libraries Version 24.04

RakeshwarK:
Added two parameters "PerformanceLibrary" and "PerformanceLibraryVersion".

For this we are breaking backward compatibility at two places
1. Removing "UsePerformanceLibraries" and set PerformanceLibrary default to null.
2. Currently the Perf Lib pkg contained "arm-performance-libraries_23.04.1_Ubuntu-22.04.sh"
   Removing the ubuntu part at the end and adding version 24.04. The 1.1.0 version pkg contains "arm-performance-libraries_23.04.1" and "arm-performance-libraries_24.04"


yangpanMS:
Could you help update the document on which versions and scripts are availble? We can't expect other people to know the combinations

yangpanMS:
Hi @rakeshwark I see you still haven't added the document for the supported blob names. Do you still need this PR?

RakeshwarK:
> Hi @RakeshwarK I see you still haven't added the document for the supported blob names. Do you still need this PR?
Hi @yangpanMS I have updated the profile and hplinpack-profiles.md. I also added a new package to Public Blob storage. Please let me know if I missed something. Yes, we need this PR, I am working on adding Intel and AMD performance libraries.

yangpanMS (approved):

PR #348, opened 2024-07-22
Correcting a specific instance of nested parameter evaluation.

mittaltarkik:
While using the Turney function, I encountered an issue. If a parameter's value is used before its evaluation, it fails with the error CS1002: ; expected. This happens because we are performing both EvaluateParameterSpecificExpressions and EvaluateWellKnownExpressionsAsync in the same loop.

To resolve this, I have moved the EvaluateParameterSpecificExpressions to a separate loop, and now this case works fine.


src/VirtualClient/VirtualClient.Core/ProfileExpressionEvaluator.cs
```diff
@@ -613,6 +613,13 @@ public async Task EvaluateAsync(IServiceCollection dependencies, IDictionary<str
                 ProfileExpressionEvaluator.EvaluateParameterSpecificExpressions(dependencies, parameters, cancellationToken);
                 await ProfileExpressionEvaluator.EvaluateWellKnownExpressionsAsync(dependencies, parameters, cancellationToken);
             }
+
```
nmalkapuram, line 616: How is this different from just increasing maxIterations for evaluation?

yangpanMS (approved):

PR #353, opened 2024-08-01
Openssl Metric Parser Fix

RakeshwarK:
The output format for RSA results changed with OpenSSL version 3.2.0.
For this fix, I have updated the parser such that it maps the metric name and its value and stores it in a list, this can parse both signverify and op results. This implementation also has a limitation, the regex references 'bits' and looks the line before that for Metric Names, so in the future if this pattern changes, we might have errors in the data that's being parsed.


src/VirtualClient/VirtualClient.Actions.UnitTests/Examples/OpenSSL/OpenSSL-speed-results-with-anomalies.txt
yangpanMS, line 1: Please do not change the previous files for backwards compatibility. Add new tests and new files for the new format

PR #355, opened 2024-08-02
Handling corner case for MaxClients parameter 

RakeshwarK:
Handling a case where we use default value for MaxClients.


src/VirtualClient/VirtualClient.Actions/Memtier/MemtierBenchmarkClientExecutor.cs
```diff
@@ -464,16 +464,22 @@ private Task ExecuteWorkloadsAsync(IPAddress serverIPAddress, ServerState server
                     int serverprocesscount = serverState.Ports.Count();
                     CpuInfo cpuInfo = this.SystemManagement.GetCpuInfoAsync(CancellationToken.None).GetAwaiter().GetResult();
                     int logicalProcessorCount = cpuInfo.LogicalProcessorCount;
+                    int maxClients = this.MaxClients;
                     int memtierProcessesCount = 0;
                     int memtierCpuAffinity = 0;
 
+                    if (maxClients == int.MaxValue)
```
yangpanMS, line 471: This looks like tribal knowledge. If you want to use max clients you need to put in int.MaxValue? Could you explain why we need this? Also wouldn't this create int overflow?

RakeshwarK: We want MaxClients to be an optional parameter. int.MaxValue is the default value we use but as we don't want to trigger int.Max number of clients, we would trigger serverinstances*clientinstances number of clients ( to support backward compatilibility)

RakeshwarK: Changed the default to -1

yangpanMS (approved):

PR #362, opened 2024-08-13
Fixing atop parser code to not consider first atop sample and output sum if counter represents number.

bhagyeshpatil:
The first atop output sample represents the activity from boot till atop execution, this duration doesn't overlap with the execution of Virtual client workloads. If the numbers are considered while emitting the counters, this will skew the results. In this PR, the atop parser is updated to not consider the first atop output sample and output the sum of individual values if the counter represents a number, e.g. in case of counter "\Disk\# Reads" the value should sum and not average of individual values.


src/VirtualClient/VirtualClient.Monitors.UnitTests/AtopParserUnitTests.cs
```diff
@@ -629,7 +620,7 @@ public void AtopParserSupportsDefiningExplicitSubsetsOfCounters_Scenario1()
             AtopParser testParser = new AtopParser(rawText);
             IList<Metric> metrics = testParser.Parse();
 
-            Assert.AreEqual(158, metrics.Count);
+            Assert.AreEqual(150, metrics.Count);
```
yangpanMS, line 623: What's the reason for the 8 less metrics? What did we stop parsing?

src/VirtualClient/VirtualClient.Monitors.UnitTests/AtopParserUnitTests.cs
```diff
@@ -13,6 +13,7 @@ namespace VirtualClient.Monitors
     using NUnit.Framework;
     using VirtualClient.Common;
     using VirtualClient.Contracts;
+    using YamlDotNet.Core.Tokens;
```
yangpanMS, line 16: Did we need this?

src/VirtualClient/VirtualClient.Monitors.UnitTests/AtopParserUnitTests.cs
bhagyeshpatil: This PR attempts to fix the first record containing a sum of all counters since system boot or counter rollover.  atop does not display any date for several of the counters because during subsequent 1s snapshots the counters remain at 0.  This eliminates records for sdb, sdc in the disk portion (busy, read, write, and avio) 2 x 4 = 8.  It does not affect the inputs for networking for the 60s example provided but potentially any network interface may not be repeated.  Something to watch out for if additional canned input is captured in the future.

src/VirtualClient/VirtualClient.Monitors.UnitTests/AtopParserUnitTests.cs
bhagyeshpatil: I don't think so, I will remove it.

yangpanMS (approved):

yangpanMS (approved):

PR #369, opened 2024-08-26
Superbenchmark executor: online inference config file and custom repo link

LiYueqian-James:
Added support for
1. Feeding an online inference config file as a link, without the need to zip into packages and upload to storage account
2. Feeding a different superbenchmark repo link other the default microsoft SB repo. Some users have added more benchmark to their own version of superbench and can use this method run them in VC.

### DIFF
--- a/src/VirtualClient/VirtualClient.Actions/SuperBenchmark/SuperBenchmarkExecutor.cs
+++ b/src/VirtualClient/VirtualClient.Actions/SuperBenchmark/SuperBenchmarkExecutor.cs
@@ -28,6 +28,7 @@ namespace VirtualClient.Actions
         private const string SuperBenchmarkRunShell = "RunSuperBenchmark.sh";
         private const string DefaultSBRepoLink = "https://github.com/microsoft/superbenchmark";
         private string configFileFullPath;
+        private string repositoryName;
 
         private IFileSystem fileSystem;
         private IPackageManager packageManager;
@@ -119,7 +120,7 @@ namespace VirtualClient.Actions
         {
             get
             {
-                return this.PlatformSpecifics.Combine(this.PlatformSpecifics.PackagesDirectory, "superbenchmark");
+                return this.PlatformSpecifics.Combine(this.PlatformSpecifics.PackagesDirectory, this.repositoryName);
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/VirtualClient/VirtualClient.Actions/SuperBenchmark/SuperBenchmarkExecutor.cs
```diff
@@ -82,6 +85,17 @@ public string ConfigurationFile
             }
         }
 
+        /// <summary>
+        /// Link to the superbench repo.
+        /// </summary>
+        public string RepositoryLink
```
yangpanMS, line 91: Do we need to change the repository link? Even if we need to support extra config?

src/VirtualClient/VirtualClient.Actions/SuperBenchmark/SuperBenchmarkExecutor.cs
```diff
@@ -151,18 +165,47 @@ protected override async Task ExecuteAsync(EventContext telemetryContext, Cancel
...
+            this.repositoryName = Path.GetFileName(repositoryLinkUri.AbsolutePath);
+            
+            // download config file if a link is provided
+            if (this.ConfigurationFile.StartsWith("http"))
+            {                
+                var configFileUri = new Uri(this.ConfigurationFile);
+                string configFileName = Path.GetFileName(configFileUri.AbsolutePath);
+                string configFullPath = this.PlatformSpecifics.Combine(this.SuperBenchmarkDirectory, configFileName);
+
+                using (var client = new HttpClient())
```
yangpanMS, line 178: Let's not add un-unit testable code here.

yangpanMS:
Could we do this instead: you are allowed to use extra dependency (for example git clone) in the profile, and you give it a package name. You then could reference the file you cloned by path.

LiYueqian-James:
> Could we do this instead: you are allowed to use extra dependency (for example git clone) in the profile, and you give it a package name. You then could reference the file you cloned by path.

so it's not just about inference config file. some people have created their own SB benchmarks (in their own repo) and want to run those. The official one do not have these benchmarks. That's why I added a parameter for repoLink

src/VirtualClient/VirtualClient.Actions/SuperBenchmark/SuperBenchmarkExecutor.cs
LiYueqian-James: Are cmd line based methods preferred? like wget or curl?

PR #376, opened 2024-09-09
Fixed Bug With Nvidia SMI Failing On NA

sasanka-grs-work:
Made changes to nvidia-smi GPU parser to support N/A values. If any metric returns N/A value, that specific metric is skipped from being logged.

image


sasanka-grs-work:

yangpanMS (approved):

PR #386, opened 2024-09-30
Adding Iteration count for LspciMonitors

deep1712:


src/VirtualClient/VirtualClient.Monitors/Lspci/LspciMonitor.cs
```diff
@@ -71,7 +71,8 @@ private async Task ListPciAsync(EventContext telemetryContext, CancellationToken
             await Task.Delay(this.MonitorWarmupPeriod, cancellationToken);
 
             bool firstRun = true;
-            while (!cancellationToken.IsCancellationRequested)
+            int iteration = 0;
+            while (!cancellationToken.IsCancellationRequested && iteration < this.MonitorIterations)
```
yangpanMS, line 75: The default of this.MonitorIterations is null, I think this will nullref

deep1712: Thanks, I had set default to -1 for infinite loop

yangpanMS (approved):

PR #390, opened 2024-10-22
adding nvlink throughput to nvidia smi monitor

ericavella:


yangpanMS:
Could we update the document for this? Like which systems should use which monitor and which profiles would be affected

src/VirtualClient/VirtualClient.Monitors/Nvidia-Smi/NvidiaSmiMonitor.cs
```diff
@@ -64,21 +82,53 @@ protected override void Validate()
...
+                        IList<Metric> metrics = parser.Parse();
+
+                        if (metrics?.Any() == true)
+                        {
+                            this.Logger.LogPerformanceCounters("nvidia", metrics, startTime, endTime, telemetryContext);
+                        }
+                    }
+                }
+            }
+            catch (OperationCanceledException)
```
yangpanMS, line 112: Could we reduce the duplicated codes here all these are calling commands and error handling, only thing that changes are command arguments and parser

yangpanMS (changes requested):
Good changes and we definitely need them. Just try to reduce the dup code and update the doc

PR #395, opened 2024-11-06
modified ProfileExpressionEvaluator

shobha-05:
Added new feature in ProfileExpressionEvaluator "IsArm" which will return "true" if the machine is arm otherwise it will return "false".

### DIFF
--- a/src/VirtualClient/VirtualClient.Core/ProfileExpressionEvaluator.cs
+++ b/src/VirtualClient/VirtualClient.Core/ProfileExpressionEvaluator.cs
@@ -5,6 +5,7 @@ namespace VirtualClient
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics.Metrics;
     using System.Linq;
     using System.Text.RegularExpressions;
     using System.Threading;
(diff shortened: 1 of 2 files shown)


src/VirtualClient/VirtualClient.Core/ProfileExpressionEvaluator.cs
```diff
@@ -104,10 +105,17 @@ public class ProfileExpressionEvaluator : IExpressionEvaluator
             @"\{([a-z0-9_-]+)\.(TotalDays|TotalHours|TotalMilliseconds|TotalMinutes|TotalSeconds)\}",
             RegexOptions.Compiled | RegexOptions.IgnoreCase);
 
+        // e.g.
+        // {Isarm}
+        private static readonly Regex Isarmexpression = new Regex(
```
yangpanMS, line 110: Let's not do IsARM. do {platform}==linux-arm64 |  {platform}==win-arm64 instead

yangpanMS (changes requested):
We already have {platform} supported to do this

shobha-05:
not required.

PR #401, opened 2024-11-13
MLPerf developer guide

saibulusu:
Will add more details on other config versions & benchmarks as support for those is added.


website/docs/workloads/mlperf/mlperf-developer-guide.md
```diff
@@ -0,0 +1,221 @@
...
+Docker is installed with DockerInstallation.
+```
+{
+    "Type": "DockerInstallation",
+    "Parameters": {
+        "Scenario": "InstallDocker"
+    }
+}
+```
+- **NVIDIA Container Toolkit**: A set of tools which enable the use of NVIDIA GPUs within docker containers.
```
yangpanMS, line 149: thanks for the explanation for all the dependencies

website/docs/workloads/mlperf/mlperf-developer-guide.md
```diff
@@ -0,0 +1,221 @@
...
+}
+```
+
+## Running a Benchmark
+There are a few setup steps before running the benchmark:
+- **make prebuild**: Download the docker container image and launch the container.
+The remaining commands are run within the container. In order to avoid launching the docker container shell,
+the file is replaced with Virtual Client. The [replacing Makefile.docker file](https://github.com/microsoft/VirtualClient/blob/main/src/VirtualClient/VirtualClient.Actions/MLPerf/Makefile.docker)
+does not launch the docker container shell.
+- **make download_data BENCHMARKS="bert"**: Download datasets necessary to run the benchmark.
```
yangpanMS, line 166: In the previous section we can mention we also need make and gcc? it's just embedded the nvidia driver install

website/docs/workloads/mlperf/mlperf-developer-guide.md
```diff
@@ -0,0 +1,221 @@
...
+- **make download_model BENCHMARKS="bert"**: Download pre-trained model to be tested.
+- **make preprocess_data BENCHMARKS="bert"**: Formats data to be used in the benchmark.
+- **make build**: Compile and build the executable to run the benchmark.
+
+To actually run the benchmark:
+- **make run RUN_ARGS='--benchmarks=bert --scenarios=Offline,Server,SingleStream --config_ver=default --test_mode=PerformanceOnly --fast**: Run performance mode which focuses
```
yangpanMS, line 172: Would be great to mention what each of these arg mean or the link to the doc to explain that.

yangpanMS (approved):

website/docs/workloads/mlperf/mlperf-developer-guide.md
saibulusu: I stated what each argument is now. Most of these are explained a little more throughout the developer guide.

yangpanMS (approved):

PR #407, opened 2024-11-27
Added optional parameters "Threads" and "Copies" to SpecCpu.

RakeshwarK:
Allowing the user to override these parameters from the Profile, with the default set to ProcessorCount for backward compatibility.


yangpanMS (approved):
Please also make changes to the INTSPEED and FPSPEED

brdeyo (approved):

PR #412, opened 2024-12-11
Add metric verbosity and allow to print critical metrics on console

yangpanMS:


src/VirtualClient/VirtualClient.Contracts/Parser/MetricsParser.cs
```diff
@@ -20,5 +22,10 @@ public MetricsParser(string rawText)
             : base(rawText)
         {
         }
+
+        /// <summary>
+        /// Metrics list
+        /// </summary>
+        public List<Metric> Metrics { get; set; } = new List<Metric>();
```
brdeyo, line 29: Yang, why do we need this? The MetricsParser already returns the set of metrics. This seems like overkill

src/VirtualClient/VirtualClient.Contracts/Enumerations.cs
```diff
@@ -677,6 +677,27 @@ public enum MetricRelativity
         LowerIsBetter = 2
     }
 
+    /// <summary>
+    /// Metric verbosity to mark the importance of a metric.
+    /// </summary>
+    public enum MetricVerbosity
```
brdeyo, line 683: I would prefer to use the existing LogLevel enumeration so that we don't need to reinvent the wheel here. It has all of the terms we would need.

e.g.
Verbosity=Trace
Verbosity=0

Verbosity=Information
Verbosity=2

Additionally, the string representation as well as the numeric representation of the enum should work.

src/VirtualClient/VirtualClient.Contracts/Extensions/MetricExtensions.cs
```diff
@@ -0,0 +1,188 @@
...
+            return filteredMetrics;
+        }
+
+        /// <summary>
+        /// Print metrics to console. Default to critical only.
+        /// </summary>
+        /// <param name="metrics">List of Metrics</param>
+        /// <param name="scenario">Scenario</param>
+        /// <param name="criticalOnly">Boolean to note if prints critical only metric</param>
+        public static void PrintConsole(this IEnumerable<Metric> metrics, string scenario, bool criticalOnly = true)
```
brdeyo, line 90: Naming: LogConsole

src/VirtualClient/VirtualClient.Contracts/Extensions/MetricExtensions.cs
```diff
@@ -0,0 +1,188 @@
...
+                // Don't throw on print console errors
+                Console.WriteLine($"Failed to print metrics on console: '{exc.Message}'");
+            }
+        }
+
+        /// <summary>
+        /// Data table with formated column width.
+        /// </summary>
+        /// <param name="dataTable">Input data table.</param>
+        private static void PrintDataTableFormatted(DataTable dataTable)
```
brdeyo, line 128: Naming: LogConsole

We don't need to be so explicit in the names of methods that are extensions and that have different input parameters. It adds unnecessary layers of abstraction.

src/VirtualClient/VirtualClient.Contracts/Extensions/MetricExtensions.cs
```diff
@@ -0,0 +1,188 @@
...
+                maxLabelSize = Math.Max(col.ColumnName.Length, maxLabelSize);
+
+                colWidths.Add(col.ColumnName, maxLabelSize);
+                for (int i = 0; i < maxLabelSize - col.ColumnName.Length + 1; i++)
+                {
+                    Console.Write(" ");
+                }
+            }
+
+            Console.Write("|");
```
brdeyo, line 149: Use ConsoleLogger.Default vs. Console.

Apply to all

src/VirtualClient/VirtualClient.Contracts/Metric.cs
```diff
@@ -126,6 +136,11 @@ public Metric(string name, double value, string unit, MetricRelativity relativit
         /// </summary>
         public IDictionary<string, IConvertible> Metadata { get; }
 
+        /// <summary>
+        /// Metric verbosity to descript importance of metric.
+        /// </summary>
+        public MetricVerbosity Verbosity { get; set; } = MetricVerbosity.Standard;
```
brdeyo, line 142: Same as noted earlier. Let's use the out of box enum we use otherwise for describing logging level/severity: LogLevel. We don't need an extra enum abstraction.

brdeyo (changes requested):

src/VirtualClient/VirtualClient.Contracts/Extensions/MetricExtensions.cs
yangpanMS: renamed

src/VirtualClient/VirtualClient.Contracts/Enumerations.cs
yangpanMS: Switched to loglevel

src/VirtualClient/VirtualClient.Contracts/Extensions/MetricExtensions.cs
yangpanMS: Won't change as discussed as console logger only prints line by line

brdeyo (approved):

PR #417, opened 2024-12-18
Added AzLinux-dnf support for RedisPackageInstallation

RakeshwarK:


src/VirtualClient/VirtualClient.Dependencies/RedisPackageInstallation.cs
```diff
@@ -89,12 +89,15 @@ protected override async Task InitializeAsync(EventContext telemetryContext, Can
             if (this.Platform == PlatformID.Unix)
             {
                 LinuxDistributionInfo distroInfo = await this.systemManager.GetLinuxDistributionAsync(cancellationToken);
+                this.Logger.LogMessage($"Print Distro Info:{distroInfo}", telemetryContext);
```
yangpanMS, line 92: Please do me a favor and set the [SupportedPlatforms("linux-arm64,linux-x64", throwError: true)] and then remove the check on line 84, line 89

src/VirtualClient/VirtualClient.Main/profiles/GET-STARTED-REDIS.json
```diff
@@ -55,22 +55,10 @@
             }
         },
         {
-            "Type": "WgetPackageInstallation",
```
yangpanMS, line 58: Have you tested the changes in ubuntu?

src/VirtualClient/VirtualClient.Dependencies/RedisPackageInstallation.cs
```diff
@@ -154,5 +160,35 @@ await this.systemManager.PackageManager.RegisterPackageAsync(redisPackage, cance
...
+        private async Task InstallOnAzLinuxAsync(EventContext telemetryContext, CancellationToken cancellationToken)
+        {
+            if (!string.IsNullOrEmpty(this.Version))
+            {
+                this.installRedisCommand = $"install redis-{this.Version} -y";
+            }
+            else
+            {
+                this.installRedisCommand = $"install redis -y";
+
```
yangpanMS, line 173: nit: blank line

src/VirtualClient/VirtualClient.Dependencies/RedisPackageInstallation.cs
```diff
@@ -154,5 +160,35 @@ await this.systemManager.PackageManager.RegisterPackageAsync(redisPackage, cance
                             .ConfigureAwait(false);
 
         }
+
+        private async Task InstallOnAzLinuxAsync(EventContext telemetryContext, CancellationToken cancellationToken)
```
yangpanMS, line 164: have you tried tdnf?

yangpanMS, line 164: I believe this would also work on RHEL/Fedora distros?

RakeshwarK: tdnf also works, is it preferred!?

RakeshwarK: Yes, but I haven't tested these cases. Should I also include RHEL/Fedora distros?

src/VirtualClient/VirtualClient.Main/profiles/GET-STARTED-REDIS.json
RakeshwarK: There seems to be some issue with Redis in ubuntu even without these changes, but the changes were running fine on AzLinux.

src/VirtualClient/VirtualClient.Dependencies/RedisPackageInstallation.cs
yangpanMS: on AzLinux yes, on RHEL still use dnf. You don;t have to change it right now, you could add a comment that tdnf also works. I heard AzLinux eventually wants to move to tdnf entirely

yangpanMS: I think so, Redis is important enough that we need more than ubuntu and AzLinux

PR #423, opened 2025-01-16
azlinux3 dnf packages for compiler installation

saibulusu:
New logic for compiler installation:
azlinux:
&nbsp; version must not be supplied
ubuntu/debian/centos/rhel:
&nbsp; if version is supplied:
&nbsp; &nbsp; install & use supplied version
&nbsp; else:
&nbsp; &nbsp; if version does not exist on system:
&nbsp; &nbsp; &nbsp; install default, then set priority

Tested on azlinux3, jammy.
Further testing for other operating systems to be done in QoS.


yangpanMS:
If we do this, are we still checking if "CompilerVersion" matches what actually got installed?. Otherwise user might want gcc14 but got gcc11 instead.

yangpanMS:
On AzLinux since you can;t specify, user need to either overwrite the compilerVersion to "", or provide a version that matches the version installed

yangpanMS:
Actually, let's do this on all other distro as well. Default compiler version to string.empty, and if that happens, install version-less gcc/gfortran etc. If they specify a version, you install gcc-10 etc. On AzLinux, we throw error if user tries to pass in a version

saibulusu:
> Actually, let's do this on all other distro as well. Default compiler version to string.empty, and if that happens, install version-less gcc/gfortran etc. If they specify a version, you install gcc-10 etc. On AzLinux, we throw error if user tries to pass in a version

How should we handle SetGccPriorityAsync when the user does not supply a version?

saibulusu:
Tested LMBench on Ubuntu, CentOS, RHEL, & AzLinux.

saibulusu:

saibulusu:
@microsoft-github-policy-service rerun

yangpanMS:
Please also test on SPECcpu, which has more complexity in compiler definition.

src/VirtualClient/VirtualClient.Dependencies/CompilerInstallation.cs
```diff
@@ -209,26 +208,59 @@ private Task InstallCygwinAsync(DependencyPath cygwinInstallationPath, EventCont
...
+                        await this.ExecuteCommandAsync("apt", @$"install build-essential gcc-{gccVersion} g++-{gccVersion} gfortran-{gccVersion} -y --quiet", Environment.CurrentDirectory, telemetryContext, cancellationToken);
+                        await this.SetGccPriorityAsync(gccVersion, telemetryContext, cancellationToken);
+                    }
 
                     break;
 
                 case LinuxDistribution.CentOS8:
                 case LinuxDistribution.RHEL8:
+                    await this.RemoveAlternativesAsync(telemetryContext, cancellationToken);
+                    if (string.IsNullOrEmpty(gccVersion))
```
yangpanMS, line 235: we also need to test for overwriting scenario: if you already have gcc-14 for example, does installing a default gcc(11 for example) sets the priority proprly by apt

yangpanMS:
Please confirm that we will send the compiler version in telemetry properly. Ideally in metrics

src/VirtualClient/VirtualClient.Dependencies/CompilerInstallation.cs
saibulusu: Should we check the compiler version before RemoveAlternativesAsync and ensure we install that one instead of the default one then? (in this example install 14)

saibulusu:
For speccpu:
Ran speccpu fprate on an azlinux3 VM, metrics can be viewed here:
```
cluster('azurecrcworkloads.westus2.kusto.windows.net').database('WorkloadPerformance').Metrics_Dev01
| where Timestamp > ago (3d)
| where ExperimentId == "bacc1a9c-543d-4a1b-9bb4-8af8af681bc5"
| where ToolName != "atop"
```
Removed gcc limit in documentation.

For Alex's query on compiler version:
I am reading from metadata to get the compiler version for speccpu when the version is not supplied by the user. So, I believe it is being correctly logged:
```
compilerMetadata.TryGetValue("compilerVersion_gcc", out compilerVersion)
```

src/VirtualClient/VirtualClient.Actions/SPECcpu/SpecCpuExecutor.cs
```diff
@@ -432,16 +433,51 @@ private async Task WriteSpecCpuConfigAsync(CancellationToken cancellationToken)
...
+                Convert.ToInt32(gccVersion) >= 10 ? SpecCpuConfigPlaceHolder.Gcc10WorkaroundContent : string.Empty,
                 StringComparison.OrdinalIgnoreCase);
 
             await this.fileSystem.File.WriteAllTextAsync(this.Combine(this.PackageDirectory, "config", configurationFile), templateText, cancellationToken);
         }
 
+        private async Task<string> GetInstalledGCCVersionAsync(CancellationToken cancellationToken)
+        {
+            string command = "gcc";
+            string commandArguments = "-dumpversion";
```
yangpanMS, line 464: speccpu uses gcc cc g++ gfortran, we need to check all of them

saibulusu: Changed to allow any compiler as a parameter.

src/VirtualClient/VirtualClient.Main/profiles/PERF-SPECCPU-FPRATE.json
yangpanMS, line 1: Also remove the compiler name and version from all the speccpu profiles, and coremark+any profile that requires compiler

src/VirtualClient/VirtualClient.Dependencies/CompilerInstallation.cs
```diff
@@ -334,6 +295,32 @@ private Task ExecuteCommandAsync(string pathToExe, string commandLineArguments,
...
+
+                        return process.StandardOutput.ToString().Trim().Split(".")[0];
+                    }
+                }
+                catch
+                {
+                }
+            }
+
+            return string.Empty;
```
yangpanMS, line 322: I assume this is not complete?

src/VirtualClient/VirtualClient.Main/profiles/PERF-SPECCPU-FPRATE.json
saibulusu: Removed compiler name. Set compiler version to an empty string. Will have to do this for the internal repo after this PR gets merged.

src/VirtualClient/VirtualClient.Actions/SPECcpu/SpecCpuExecutor.cs
saibulusu: Changed this again, only checking for gcc version in speccpu now.

src/VirtualClient/VirtualClient.Dependencies/CompilerInstallation.cs
saibulusu: Updated this method. It gets the major version using '-dumpversion'.

yangpanMS (approved):
Approved but please update the documement for CompilerInstallation on the new behavior

yangpanMS (approved):

PR #426, opened 2025-01-22
Add conditional -k switch to CPS client command line

Lekshmi-Ra:
This switch is required to insert 2 seconds delay between send/receive when the data transfer mode is set to 2, which is continuous send/receive.

### DIFF
--- a/src/VirtualClient/VirtualClient.Actions/Network/NetworkingWorkload/CPS/CPSClientExecutor.cs
+++ b/src/VirtualClient/VirtualClient.Actions/Network/NetworkingWorkload/CPS/CPSClientExecutor.cs
@@ -41,7 +41,7 @@ namespace VirtualClient.Actions.NetworkPerformance
             string clientIPAddress = this.GetLayoutClientInstances(ClientRole.Client).First().IPAddress;
             string serverIPAddress = this.GetLayoutClientInstances(ClientRole.Server).First().IPAddress;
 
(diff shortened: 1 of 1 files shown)


src/VirtualClient/VirtualClient.Actions/Network/NetworkingWorkload/CPS/CPSClientExecutor.cs
```diff
@@ -41,7 +41,7 @@ protected override string GetCommandLineArguments()
             string clientIPAddress = this.GetLayoutClientInstances(ClientRole.Client).First().IPAddress;
             string serverIPAddress = this.GetLayoutClientInstances(ClientRole.Server).First().IPAddress;
 
-            return $"-c -r {this.Connections} " +
+            return $"-c -r {this.Connections} " +  $"{((this.DataTransferMode == 2) ? "-k " : string.Empty)}" +
```
yangpanMS, line 44: You will want to add unit test for this case

brdeyo (changes requested):
Please update the VERSION file at the root of the repo. This ensures that a new NuGet package is published to Nuget.org. This is required for us to bring these changes into the internal builds/releases of VC.

Lekshmi-Ra:
Abandoning this PR as we're working on another approach which adds more flexibility to the execution of the tool.

PR #436, opened 2025-02-12
Returning charm++ as a compiler

saibulusu:
Undoing part of the change in the [previous commit](https://github.com/microsoft/VirtualClient/commit/8b066011e97a74fc3a17a6d802218c90c80f51de) to bring back charm++ as a compiler. CompilerName is an optional parameter and will default to 'gnu', which refers to gcc, cc, gfortran, etc.


src/VirtualClient/VirtualClient.Dependencies/CompilerInstallation.cs
```diff
@@ -41,6 +42,22 @@ public CompilerInstallation(IServiceCollection dependencies, IDictionary<string,
...
+        /// The name of the compiler (e.g. gnu).
+        /// </summary>
+        public string CompilerName
+        {
+            get
+            {
+                return this.Parameters.GetValue<string>(nameof(CompilerInstallation.CompilerName), "gnu");
+            }
+
+            set
```
yangpanMS, line 55: we don't need the set

yangpanMS (approved):

yangpanMS (approved):

PR #439, opened 2025-02-13
Log Process Details in ExecuteCommandAsync

ericavella:


src/VirtualClient/VirtualClient.Core/VirtualClientComponentExtensions.cs
```diff
@@ -153,6 +153,9 @@ public static async Task<IProcessProxy> ExecuteCommandAsync(
                 beforeExecution?.Invoke(process);
                 await process.StartAndWaitAsync(cancellationToken)
                     .ConfigureAwait(false);
+
+                await component.LogProcessDetailsAsync(process, telemetryContext, logToFile: true);
```
yangpanMS, line 157: Please use the LogToFile on the component

src/VirtualClient/VirtualClient.Core/VirtualClientComponentExtensions.cs
```diff
@@ -153,6 +153,9 @@ public static async Task<IProcessProxy> ExecuteCommandAsync(
...
+                process.ThrowIfErrored<ProcessException>(process.StandardError.ToString(), ErrorReason.ProcessFailed);
```
yangpanMS, line 158: Let's not throw here for now, people usually handle it outside

yangpanMS (approved):
Thank you for cleaning this up. Would you mind picking a couple older workloads that still have private async Task ExecuteCommandAsync and see if you could remove that 20 lines of code and use this instead?

PR #440, opened 2025-02-13
Retrieving Disk Paths via DiskManager

ericavella:


src/VirtualClient/VirtualClient.Core/UnixDiskManager.cs
```diff
@@ -169,6 +169,45 @@ public override Task<IEnumerable<Disk>> GetDisksAsync(CancellationToken cancella
...
+        /// </summary>
+        /// <param name="cancellationToken"></param>
+        /// <param name="diskFilter">Disk filter to </param>
+        /// <returns></returns>
+        /// <exception cref="WorkloadException"></exception>
+        public override Task<IEnumerable<string>> GetDiskPathsAsync(string diskFilter, CancellationToken cancellationToken)
+        {
+            List<string> diskPaths = new List<string>();
+
+            return this.Logger.LogMessageAsync($"{nameof(UnixDiskManager)}.GetDiskPaths", EventContext.Persisted(), async () =>
```
yangpanMS, line 183: You only need this wrap if you care about start and ends, do you need that information?

src/VirtualClient/VirtualClient.Core/UnixDiskManager.cs
```diff
@@ -169,6 +169,45 @@ public override Task<IEnumerable<Disk>> GetDisksAsync(CancellationToken cancella
...
+                if (disksToTest?.Any() != true)
+                {
+                    throw new WorkloadException(
+                        "Expected disks to test not found. Given the parameters defined for the profile action/step or those passed " +
+                        "in on the command line, the requisite disks do not exist on the system or could not be identified based on the properties " +
+                        "of the existing disks.",
+                        ErrorReason.DependencyNotFound);
+                }
+
+                foreach (Disk disk in disksToTest)
```
yangpanMS, line 197: If I'm understanding correctly, you just need the extension methods on Disks -> Disk.GetPreferredAccessPath? You don't need to put that into DiskManager if that's the case

yangpanMS (changes requested):
I need to understand more about this. Could you use Disks.GetPreferredPath in the component that needs this? If you don't want to use foreach, you could write an extension just on the Disks.

src/VirtualClient/VirtualClient.Core/UnixDiskManager.cs
ericavella: GetPreferredAccessPath is leveraged to get the access paths of all the disks attached / filtered by DiskFilter. Since this method appears in >2 workloads & dependencies, thought it would be appropriate as a method through the disk manager

yangpanMS: If that's the case you should call the GetDisks function above to avoid the duplicated filter code

src/VirtualClient/VirtualClient.Actions/FIO/FioExecutor.cs
```diff
@@ -313,37 +315,15 @@ protected override async Task ExecuteAsync(EventContext telemetryContext, Cancel
...
-                    if (disksToTest?.Any() != true)
-                    {
-                        throw new WorkloadException(
-                            "Expected disks to test not found. Given the parameters defined for the profile action/step or those passed " +
-                            "in on the command line, the requisite disks do not exist on the system or could not be identified based on the properties " +
-                            "of the existing disks.",
-                            ErrorReason.DependencyNotFound);
-                    }
-
-                    if (await this.CreateMountPointsAsync(disksToTest, cancellationToken).ConfigureAwait(false))
```
yangpanMS, line 327: In FIO we still need to create mount points,

ericavella: I changed the profile to reflect that

yangpanMS:
Could you remind me if we still need this? Also would we change this for FIO only or we need to modify it for all IO workloads

PR #442, opened 2025-02-16
Fixing fio job file error

saibulusu:
Observing this error in QoS VC for PERF-IO-FIO-OLTP.json:
One or more errors occurred. (Maximum Virtual Client restart-on-crash attempts reached (crashes = 6). stderr: [02/15/2025 09:10:19] Unexpected profile definition. Only one of JobFiles or CommandLine can be defined.\n\n   at VirtualClient.Actions.FioExecutor.Validate()\n   at VirtualClient.Contracts.VirtualClientComponent.<>c__DisplayClass130_1.<<ExecuteAsync>b__4>d.MoveNext()\n--- End of stack trace from previous location ---\n ...

This seems to occur on the second iteration of the workload during parameter validation. Changing the job file case to avoid setting the command line in that case.
Tested PERF-IO-FIO-OLTP and PERF-IO-FIO manually on VMs.

### DIFF
--- a/src/VirtualClient/VirtualClient.Actions/FIO/FioExecutor.cs
+++ b/src/VirtualClient/VirtualClient.Actions/FIO/FioExecutor.cs
@@ -980,7 +980,11 @@ namespace VirtualClient.Actions
 
             foreach (string key in this.Parameters.Keys)
             {
-                text = Regex.Replace(text, @$"\${{{key.ToLower()}}}", this.Parameters.GetValue<string>(key));
+                string value = this.Parameters.GetValue<string>(key);
+                if (!string.IsNullOrEmpty(value))
+                {
+                    text = Regex.Replace(text, @$"\${{{key.ToLower()}}}", value);
+                }
             }
 
             this.SystemManagement.FileSystem.File.WriteAllText(@destinationPath, text);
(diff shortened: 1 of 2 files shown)


src/VirtualClient/VirtualClient.Actions/FIO/FioExecutor.cs
```diff
@@ -340,12 +340,18 @@ protected override async Task ExecuteAsync(EventContext telemetryContext, Cancel
...
+                    string commandLine;
+
+                    if (string.IsNullOrEmpty(this.JobFiles))
+                    {
+                        commandLine = this.CommandLine;
+                    }
+                    else
                     {
-                        await this.SetCommandLineForJobFilesAsync(cancellationToken);
+                        commandLine = await this.GetCommandForJobFilesAsync(cancellationToken);
```
yangpanMS, line 351: Could it be that during second iteration, we just need to reset the this.CommandLine?

it would still be better for use to either stick to this.CommandLine, or just use local varaible, and not in a mix

yangpanMS (approved):
Approved but please see if we could either fix the original this.CommandLine (for example clearing it every iteration, or don't set it second iteration), since that is also used in metadata and logging, or just use local variable and remove this.CommandLine

brdeyo (approved):

src/VirtualClient/VirtualClient.Actions/FIO/FioExecutor.cs
saibulusu: Using a separate variable to log command line in EvaluateParametersAsync.

yangpanMS (approved):

src/VirtualClient/VirtualClient.Actions/FIO/FioExecutor.cs
saibulusu: Using only command line now.

PR #453, opened 2025-03-02
Redis AzLinux3 profile

saibulusu:
Profile for redis on azlinux3. Used a separate profile because some commands must be executed only on azlinux3, a file must be replaced, and a certain package must be installed (pcre2-devel).
Cannot test client/server profiles on VMs I've created, unable to connect the VMs.


saibulusu:
Moving this profile to internal.

PR #454, opened 2025-03-04
mysql minor changes and sysbench cleanup addition

nmalkapuram:


src/VirtualClient/VirtualClient.Main/profiles/PERF-POSTGRESQL-SYSBENCH-OLTP.json
```diff
@@ -182,7 +182,7 @@
             "Parameters": {
                 "Scenario": "DownloadSysbenchPackage",
                 "BlobContainer": "packages",
-                "BlobName": "sysbench-1.0.20.rev2.zip",
+                "BlobName": "sysbench-1.0.20.rev3.zip",
```
yangpanMS, line 185: Is the clean up python the only change in rev3? Please put the scripts as scripts in VC source, in that way I could review the changes. Otherwise you need to document what has changed that are not in the source

src/VirtualClient/VirtualClient.Actions/Sysbench/sysbench scripts/cleanup-database.py
```diff
@@ -0,0 +1,22 @@
+import subprocess, argparse, os
```
yangpanMS, line 1: no space in directories, just scripts

src/VirtualClient/VirtualClient.Actions.UnitTests/Sysbench/SysbenchMetricsParserTests.cs
```diff
@@ -46,5 +46,17 @@ public void SysbenchParserParsesCorrectly()
             MetricAssert.Exists(metrics, "latency p95", 68.05, "milliseconds");
             MetricAssert.Exists(metrics, "latency sum", 7458385.25, "milliseconds");
         }
+
+        [Test]
+        public void SysbenchParserParsesMetricsMetadataCorrectly()
```
yangpanMS, line 51: Do you need to add an example for this? or you could just use a previous example

src/VirtualClient/VirtualClient.Actions/Sysbench/sysbench scripts/cleanup-database.py
yangpanMS, line 1: Can we put these scripts in VC build and not mysql package? This way you make sure the code and script match in the same version, you don't have situation people running rev2 in their profile with new VC

src/VirtualClient/VirtualClient.Actions.UnitTests/Sysbench/SysbenchMetricsParserTests.cs
nmalkapuram: There is a small difference between the outputs(version of sysbench is different), the parser has to work for both of them. So added a new example.

psingla1210 (approved):

PR #455, opened 2025-03-06
Fixing unit tests for PR

saibulusu:
(Work in Progress)

…p the NuGet package size below 250MB to avoid max size limitation on NuGet.org."

This reverts commit c08459b65572a0edcd32767c5e49ec88ecc080ac.


yangpanMS:
Why do we need to revert?

PR #459, opened 2025-03-07
Update Module.props to internal version

v-safilho:


saibulusu (commented):
Is this to create a nuget package? We already have a VERSION file and the package is created based on the value there.

v-safilho:
I need to let this VC version available in Juno environment
image

v-safilho:
this PR must be ignored as it has internal purposes

PR #462, opened 2025-03-11
Move delay in WindowsPerformanceMonitor SnapshotCountersAsync

moenc32:
Move "MonitorFrequency" delay in WindowsPerformanceMonitor SnapshotCountersAsync to end of while loop. This ensures that when the first round of metrics to be taken we do not have to wait for "MonitorFrequency" before generating and emitting first round of metrics to telemetry.


yangpanMS (approved):

yangpanMS (changes requested):
will sync offline

yangpanMS:
Closing as discussed

PR #464, opened 2025-03-13
Using fio from package manager

saibulusu:
Using fio from package manager instead of from storage.


src/VirtualClient/VirtualClient.Actions/FIO/FioMultiThroughputExecutor.cs
```diff
@@ -543,11 +544,16 @@ private async Task ExecuteWorkloadAsync(string testName, Dictionary<string, ICon
...
+                string updatedJobFileFolder = Path.Combine(jobFileFolder, "updated");
+
+                if (!this.SystemManagement.FileSystem.Directory.Exists(updatedJobFileFolder))
+                {
+                    this.SystemManagement.FileSystem.Directory.CreateDirectory(updatedJobFileFolder);
+                }
+
                 string templateJobFilePath = this.PlatformSpecifics.Combine(jobFileFolder, this.TemplateJobFile);
-                string jobFilePath = this.PlatformSpecifics.Combine(workloadPackage.Path, nameof(FioMultiThroughputExecutor) + this.TemplateJobFile);
+                string jobFilePath = this.PlatformSpecifics.Combine(jobFileFolder, "updated", nameof(FioMultiThroughputExecutor) + this.TemplateJobFile);
```
yangpanMS, line 556: Is it not working? why do we need to change this

yangpanMS (approved):

src/VirtualClient/VirtualClient.Actions/FIO/FioMultiThroughputExecutor.cs
saibulusu: We would have to create the directory. But if we're not using the package manager, I just kept the updated files under scripts near the template files.

yangpanMS (approved):

yangpanMS (approved):

PR #471, opened 2025-03-21
Do not run unsupported components in case of ParallelLoopExecution

imadityaa:


brdeyo (approved):

src/VirtualClient/VirtualClient.Contracts/ParallelLoopExecution.cs
```diff
@@ -68,6 +68,12 @@ private async Task ExecuteComponentLoopAsync(VirtualClientComponent component, E
         {
             while (!cancellationToken.IsCancellationRequested)
             {
+                if (!VirtualClientComponent.IsSupported(component))
```
brdeyo, line 71: Move the check for IsSupported to the "for" loop. That is a more natural place for it to be versus a check + break.

brdeyo: Please add the same check to the ParallelExecution class "for" loop as well so that we are thorough here with the feature addition.

brdeyo (changes requested):

src/VirtualClient/VirtualClient.Contracts/ParallelLoopExecution.cs
imadityaa: Sure, made the change.

yangpanMS (approved):

brdeyo (approved):

PR #478, opened 2025-04-02
Modify ThreadCount for Prime95

imadityaa:
With Hyperthread=1 and ThreadCount > Physical CoreCount, Prime95 is giving an error statement of:

Cannot initialize FFT code, errorcode 1010
Unknown gwnum error code: 1010

Either option is to set Hyperthread=0 or reduce the threadcount.

For full system stress (power, thermal and stability), hyperthread=1 and ThreadCount=PhysicalCoreCount is an optimum solution.

### DIFF
--- a/src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-PRIME95.json
+++ b/src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-PRIME95.json
@@ -7,7 +7,7 @@
     },
     "Parameters": {
         "Duration": "00:15:00",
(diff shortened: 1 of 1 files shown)


yangpanMS (approved):
When you are running on a system without hyperthreading, how much stress does it put?

imadityaa:
> When you are running on a system without hyperthreading, how much stress does it put?

If hyperthreading=0, threads=physicalCoreCount, stress is approx 50%
If hyperthreading=0, threads>=logicalCores, stress is approx 75-80%.

Full system stress (~100%) is coming up with hypertheading=1, and threads=physicalCoreCount

PR #485, opened 2025-04-09
Using a different package for speccpu

saibulusu:
Tested on speccpu fprate on ubuntu 24 VM.


yangpanMS:
As discussed, please add a comment in metadata to mention why we are using a private fix of SPECcpu 1.1.9 instead of an official release

PR #490, opened 2025-04-22
Fix parallel loop execution

rudraptpsingh:
The ParallelLoopExecution keeps on running failing component in loop, which causes unnecessary trace and log generation. Modifying the behaviour to skip running a failing component in loop

### DIFF
--- a/src/VirtualClient/VirtualClient.Contracts/ParallelLoopExecution.cs
+++ b/src/VirtualClient/VirtualClient.Contracts/ParallelLoopExecution.cs
@@ -74,26 +74,28 @@ namespace VirtualClient.Contracts
         {
             while (!cancellationToken.IsCancellationRequested)
             {
-                if (!string.IsNullOrWhiteSpace(component.Scenario))
+                try
                 {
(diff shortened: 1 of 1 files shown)


rudraptpsingh:

src/VirtualClient/VirtualClient.Contracts/ParallelLoopExecution.cs
```diff
@@ -74,26 +74,28 @@ private async Task ExecuteComponentLoopAsync(VirtualClientComponent component, E
...
+                    {
+                        break;
+                    }
+
+                    await componentExecutionTask;
                 }
-                else
+                catch (Exception ex)
                 {
+                    this.Logger.LogMessage($"{nameof(ParallelLoopExecution)} Component = {component.TypeName} encountered an error: {ex.Message}", LogLevel.Error, telemetryContext);
```
imadityaa, line 98: Instead of just logging, should we throw a non-fatal exception like "WorkloadException" if any of the workload fails? This would make sure that workload related errors are not missed or treated as false positives?

rudraptpsingh: Ack.

PR #491, opened 2025-04-23
Added Arm Perf Lib v25 support for HPLinpack Workload

RakeshwarK:
Added support for Arm Performance library version 25.04.1.


website/docs/workloads/hplinpack/hplinpack-profiles.md
```diff
@@ -42,7 +42,7 @@ This profile is designed to identify general/broad regressions when compared aga
   |   ProblemSizeN      |  The order of coefficient matrix of set of linear equations that we want to solve  | Convert.ToInt32(Math.Sqrt(totalAvailableMemoryKiloBytes * 1024 * 0.8 / 8)) (This value is dependent on memory of machine, uses 80% of available memory) |
   |   BlockSizeNB       |  The partitioning blocking factor  | 256 |
   |   PerformanceLibrary | Optional. This parameter allows you to specify machine-specific performance libraries. You can assign values such as ARM, AMD, and INTEL to utilize the corresponding performance libraries. | null |
-  | PerformanceLibraryVersion | Required when using PerformanceLibrary parameter. Specify the version of the performance libraries you would like to use.<br/><br/>Curently, the supported configurations are :<br/>ARM - 23.04.1 , 24.10 | null  |  | CCFLAGS | compiler flags| -O3 -march=native  |
+  | PerformanceLibraryVersion | Required when using PerformanceLibrary parameter. Specify the version of the performance libraries you would like to use.<br/><br/>Curently, the supported configurations are :<br/>ARM - 23.04.1, 24.10 and 25.04.1 | null  |  | CCFLAGS | compiler flags| -O3 -march=native  |
```
brdeyo, line 45: It is a bit confusing that the PerformanceLibraryVersion parameter is required while the PerformanceLibrary parameter is optional? If one of these is required, then both of them should be required so as to have cohesion in the profile.

brdeyo (changes requested):

website/docs/workloads/hplinpack/hplinpack-profiles.md
RakeshwarK: The use of Performance Libraries like 'ARM' is optional, but the perf lib version parameter is required only when we want to use Performance libraries as we support multiple versions.

To make it simple, we can have a separate profile for Performance Libraries, like PERF-CPU-HPLINPACK-PERF-LIB.json. 
This way we could also skip downloading large performance libraries when the user doesn't want to use them.
Please let me know your thoughts.

src/VirtualClient/VirtualClient.Main/profiles/PERF-CPU-HPLINPACK.json
```diff
@@ -5,14 +5,16 @@
...
-    "BindToCores": false,
-    "CCFLAGS": "-O3 -march=native"
-  },
+    "Parameters": {
+        "CompilerVersion": "",
+        "ProblemSizeN": null,
+        "BlockSizeNB": "256",
+        "NumberOfProcesses": "{LogicalCoreCount}",
+        "PerformanceLibrary": null,
+        "PerformanceLibraryVersion": "",
```
RakeshwarK, line 14: #brdeyo
I hope the minor changes made to Profile and documentation would reduce the complexity for User, till we could add support for Parameter Sets in VC Profiles.

brdeyo (approved):

PR #497, opened 2025-04-28
Compilation failed error message

saibulusu:
Added error message 513.
Tested on SFR and LMBench when gcc and make are not installed. However, the error message itself **does not** appear right now.
Clarification: should we add logging to process.ThrowIfErrored to include the error message?


src/VirtualClient/VirtualClient.Actions/HPLinpack/HPLinpackExecutor.cs
```diff
@@ -209,15 +209,24 @@ protected override async Task ExecuteAsync(EventContext telemetryContext, Cancel
...
                 DateTime startTime = DateTime.UtcNow;
-                await this.ExecuteCommandAsync("make", $"arch=Linux_GCC", this.HPLDirectory, telemetryContext, cancellationToken)
-                    .ConfigureAwait(false);
+                using (process = await this.ExecuteCommandAsync("make", $"arch=Linux_GCC", this.HPLDirectory, telemetryContext, cancellationToken).ConfigureAwait(false))
+                {
+                    if (!cancellationToken.IsCancellationRequested)
+                    {
+                        if (process.IsErrored())
+                        {
+                            await this.LogProcessDetailsAsync(process, telemetryContext, "HPLinpack", logToFile: true);
```
yangpanMS, line 221: the logProcessDetails should be outside of the process.IsErrored, also someworkloads should already be logging process details, see if there are duplications

PR #499, opened 2025-04-28
Updated ProxyBlobManager source to be public property

godsanchez:
Updated the "Source" property to allow for arbitrary blob download URLs.


yangpanMS:
As discussed we don't need this PR.

PR #504, opened 2025-05-07
Adding cool down period for specific executors

deep1712:


src/VirtualClient/VirtualClient.Contracts/VirtualClientComponent.cs
```diff
@@ -175,6 +175,17 @@ public Guid? ClientRequestId
         /// </summary>
         public ComponentType ComponentType { get; set; }
 
+        /// <summary>
+        /// The Cool down period for Virtual Client Component.
+        /// </summary>
+        public TimeSpan CoolDownPeriod
```
yangpanMS, line 181: There are currently "MinimumExecutionInterval" on the profile level to define intervals between iterations. Also if you just need it on a single component, add it in your component first instead of making it common first.

yangpanMS (changes requested):
Blocking until requirement is understood

brdeyo (approved):

yangpanMS (approved):

yangpanMS (approved):

PR #505, opened 2025-05-08
WaitForExitAsync: Handle timeout exception when timeout parameter set

rudraptpsingh:


src/VirtualClient/VirtualClient.Common/ProcessProxy.cs
```diff
@@ -365,6 +365,10 @@ public virtual async Task WaitForExitAsync(CancellationToken cancellationToken,
             {
                 // Expected whenever the CancellationToken receives a cancellation request.
             }
+            catch (TimeoutException) when (timeout != null)
```
brdeyo, line 368: This should NOT be swallowed here. There are upstream cases where the caller of this logic needs to know if the process did not actually complete within the timeout allotted. TimeoutException should be handled in the calling locations, not the core method.

brdeyo (changes requested):

PR #510, opened 2025-05-15
Adding openssl version to metadata dependencies

dheeparaj:
Ref: https://msazure.visualstudio.com/One/_workitems/edit/32307803

tested in linux platforms from VC runs - exp ID "d2ac614d-1f1a-4d72-858c-5bd099afdafa"
 "metadata_dependencies": {
    "openSSLVersion": "OpenSSL 3.0.0-beta3-dev  (Library: OpenSSL 3.0.13 30 Jan 2024)",
    "package_openssl": "openssl.3.0.0.zip"
  },


src/VirtualClient/VirtualClient.Actions/OpenSSL/OpenSslExecutor.cs
```diff
@@ -115,7 +140,8 @@ private void CaptureMetrics(IProcessProxy workloadProcess, string commandArgumen
                         null,
                         commandArguments,
                         this.Tags,
-                        telemetryContext);
+                        telemetryContext, 
+                        opensslVersion);
```
yangpanMS, line 144: Prefix with toolVersion:
I believe the first parameter after telemetry in LogMetrics is toolResults instead of toolversion

src/VirtualClient/VirtualClient.Actions/OpenSSL/OpenSslExecutor.cs
```diff
@@ -90,12 +90,37 @@ await this.InitializeWorkloadToolsetsAsync(cancellationToken)
                 .ConfigureAwait(false);
         }
 
+        private string GetOpenSslVersion()
```
yangpanMS, line 93: Ideally this should be handled in the package installation instead of here, could you explain your use case and why you need the version?

dheeparaj: The intention is to get the actual tool version from openssl with details of release date along with the major, minor, patch versions (as reported by the software). Package version is extracted from the folder naming from blob store (could be error prone possibly).

src/VirtualClient/VirtualClient.Actions/OpenSSL/OpenSslExecutor.cs
dheeparaj: thank you for noticing, fixed it.

src/VirtualClient/VirtualClient.Actions/OpenSSL/OpenSslExecutor.cs
dheeparaj: For instance, observation from one of the test i ran - 
package_name is "openssl.3.0.0.zip"
where as the actual version reported from openssl command is 
"**OpenSSL 3.0.0-beta3-dev  (Library: OpenSSL 3.0.13 30 Jan 2024)**"
please note version also display the library version number with patch number and release date. 

For better scrutiny, having the right version from the tool that is used is more accurate.

dheeparaj:
> 1. Please change the method to Async and taking cancellationToken
> 2. Please add unit test to verify the behavior

This is done

yangpanMS (approved):

PR #520, opened 2025-05-28
Users/vdeepar/update metrics

dheeparaj:
CaptureMetrics calls GetOpensslVersion which is async 
hence add await to CaptureMetrics - so it shows version update for every metric


src/VirtualClient/VirtualClient.Actions/OpenSSL/OpenSslExecutor.cs
```diff
@@ -94,7 +94,7 @@ await this.InitializeWorkloadToolsetsAsync(cancellationToken)
         /// <summary>
         /// Gets openssl version by running openssl version command.
         /// </summary>
-        private async Task GetOpenSslVersion(CancellationToken cancellationToken, string toolCommand)
+        private async Task GetOpenSslVersionAsync(CancellationToken cancellationToken, string toolCommand)
```
yangpanMS, line 97: one more nit change: put string before CancellationToken

yangpanMS (approved):

PR #525, opened 2025-06-03
Refactor logic to enable providing experiment ID and additional metadata to summary file logger.

brdeyo:
The summary file logger requires access to the experiment ID and logs folder. This refactor ensures that both of these pieces of information are available.


src/VirtualClient/VirtualClient.Contracts/VirtualClientComponent.cs
```diff
@@ -30,6 +30,12 @@ public abstract class VirtualClientComponent : IDisposable
         /// </summary>
         public static readonly char[] CommonDelimiters = new char[] { ',', ';' };
```
yangpanMS, line 31: should we merge with the previous one?

src/VirtualClient/VirtualClient.Core/Logging/SummaryFile/SummaryFileLoggerProvider.cs
```diff
@@ -27,17 +27,26 @@ public SummaryFileLoggerProvider(string filePath)
...
+
+            if (string.IsNullOrWhiteSpace(effectiveFilePath))
+            {
+                PlatformSpecifics platformSpecifics = VirtualClientRuntime.PlatformSpecifics
+                   ?? new PlatformSpecifics(Environment.OSVersion.Platform, RuntimeInformation.ProcessArchitecture);
+
+                string experimentId = VirtualClientRuntime.ExperimentId;
+                string logsPath = platformSpecifics.GetLogsPath();
+
+                effectiveFilePath = platformSpecifics.Combine(logsPath, $"{experimentId ?? DateTime.UtcNow.ToString("yyyy-MM-ddTHH-mm-ss-fff")}-summary.txt");
```
yangpanMS: We will always have an experimentId

yangpanMS (approved):

src/VirtualClient/VirtualClient.Contracts/VirtualClientComponent.cs
brdeyo: I pulled in the other changes

src/VirtualClient/VirtualClient.Core/Logging/SummaryFile/SummaryFileLoggerProvider.cs
brdeyo: We will when running as VC proper but NOT when we are running as a unit test. The logic is really just a minor workaround. If you think it unnecessary, we can just use "summary.txt" if an experiment ID is not defined.

src/VirtualClient/VirtualClient.Common/Extensions/StringExtensions.cs
```diff
@@ -51,6 +52,33 @@ public static string RemoveWhitespace(this string value)
...
+            if (text != null && values?.Any() == true)
+            {
+                for (int i = 0; i < values.Length; i++)
+                {
+                    string currentValue = values[i];
+                    index = text.IndexOf(currentValue, comparison);
+
+                    if (index >= 0)
+                    {
+                        break;
```
yangpanMS: I believe the break here is problematic for the "earliest" statement. If a string is 
A,B;C
and the delimeters are ';',','
It would've break at index 3 instead of the comma which is earlier

src/VirtualClient/VirtualClient.Core/Logging/SummaryFile/SummaryFileLoggerProvider.cs
brdeyo: I changed the logic to use the experiment ID if defined (typical VC runtime scenario).

src/VirtualClient/VirtualClient.Common/Extensions/StringExtensions.cs
brdeyo: We don't support inter-mixing the delimiters. IndexOfAny breaks at the very first occurrence of ANY of the delimiters which is by design.

brdeyo: I will add unit tests to ensure the logic is sound.

yangpanMS: no I understand, if we are not supporting mixing then it makes sense

yangpanMS (approved):

src/VirtualClient/VirtualClient.Common/Extensions/StringExtensions.cs
brdeyo: You are right. I am just going to remove the extension method. The logic we have ONLY needs to get the index of a semi-colon and not any of the three possibilities. Choosing the simpler route.

brdeyo: removed the extension and am just using IndexOf() on the first semi-colon.

yangpanMS (approved):

PR #526, opened 2025-06-04
Recording redis version in the executor

vaishalitvv:
Summary:
This update enables the recording of the Redis version on the server side using the RedisServerExecutor.

Testing:
----------------------------------------------------------------------------------------------------------------------------
// PERF-REDIS profile test for recording redis version change
cluster('azurecrcworkloads.westus2.kusto.windows.net').database('WorkloadDIagnostics').Traces_PPE01
| where ExperimentId == "875a16d5-7e08-481b-8bcc-5da542d6dd35"
| where CustomDimensions contains "RedisVersion"

// ABC-BASE-REDIS-S1_RR.json pkgmgr-based profile test for recording redis version change
cluster('azurecrcworkloads.westus2.kusto.windows.net').database('WorkloadDIagnostics').Traces_PPE01
| where ExperimentId == "50b572a4-e6d1-4abd-8b74-2d5621998c3f"
| where CustomDimensions contains "RedisVersion"

// ABC-BASE-REDIS-S1_RR-AzLinux.json wget-based profile test for recording redis version change
cluster('azurecrcworkloads.westus2.kusto.windows.net').database('WorkloadDIagnostics').Traces_PPE01
| where ExperimentId == "491fd0c1-df73-4312-941c-249cd903f898"
| where CustomDimensions contains "RedisVersion"


src/VirtualClient/VirtualClient.Actions/Redis/RedisServerExecutor.cs
```diff
@@ -264,6 +268,51 @@ protected override void Validate()
...
+        {
+            string packagePath = this.RedisPackagePath;
+            string srcDir = Path.Combine(packagePath, "src");
+            string redisServerSrc = Path.Combine(srcDir, "redis-server");
+            string redisServerDst = Path.Combine(packagePath, "redis-server");
+            try
+            {
+                if (File.Exists(redisServerSrc) && !File.Exists(redisServerDst))
+                {
+                    File.Copy(redisServerSrc, redisServerDst, overwrite: true);
```
yangpanMS: why do we need this method? also this is not unit testable

vaishalitvv: Since there are two types of Redis profiles -- ABC-BASE-REDIS-S1_RR.json (which uses Redis installed via package managers) and PERF-REDIS (which downloads Redis using wget from source) -- we need a common path that both can access. This method ensures the redis-server binary from the source-based installation is copied into the package-managed directory path so that version detection behave consistently across both profile types.
This is necessary because the path ${RedisPackagePath}/src/Redis does not exist in the package-managed installations.

RakeshwarK: This case is handled by the dependency RedisPackageInstallation.cs right!?
https://github.com/microsoft/VirtualClient/blob/main/src/VirtualClient/VirtualClient.Dependencies/RedisPackageInstallation.cs#L146

vaishalitvv: Yes, the case is being handled in the dependency RedisPackageInstallation.cs. Hence, removed this function "EnsureRedisServerInRootPackagePath"

src/VirtualClient/VirtualClient.Actions/Redis/RedisServerExecutor.cs
```diff
@@ -264,6 +268,29 @@ protected override void Validate()
...
+                this.Logger.LogMessage($"{this.TypeName}.RedisVersionCaptured", LogLevel.Information, telemetryContext);
+                this.MetadataContract.AddForScenario(
+                    "Redis-Benchmark",
+                    null,
+                    toolVersion: this.RedisVersion);
+                this.MetadataContract.Apply(telemetryContext);
+            }
+            else
+            {
+                throw new WorkloadException("Failed to parse Redis version from output.", ErrorReason.CriticalWorkloadFailure);
```
yangpanMS: Should we make this best-effort? or you this should crash entire component if we fail to get version

yangpanMS (approved):

src/VirtualClient/VirtualClient.Actions/Redis/RedisServerExecutor.cs
vaishalitvv: Changed the code to not crash the workload if it fails to process the version.

yangpanMS (approved):

PR #527, opened 2025-06-11
Added Conditional parameter sets feature

RakeshwarK:


src/VirtualClient/VirtualClient.Common/Contracts/ParameterDictionaryListJsonConverter.cs
```diff
@@ -0,0 +1,140 @@
...
+    using System.Linq;
+    using Newtonsoft.Json;
+    using Newtonsoft.Json.Linq;
+    using VirtualClient.Common.Extensions;
+
+    /// <summary>
+    /// Provides a JSON converter that can handle the serialization/deserialization of
+    /// <see cref="List{T}"/> objects where T is <see cref="IDictionary{TKey, TValue}"/> with string keys and <see cref="IConvertible"/> values.
+    /// </summary>
+    public class ParameterDictionaryListJsonConverter : JsonConverter
```
brdeyo, line 17: Naming: ParameterDictionaryCollectionJsonConverter

src/VirtualClient/VirtualClient.Core/ProfileExpressionEvaluator.cs
```diff
@@ -188,6 +210,35 @@ public class ProfileExpressionEvaluator : IExpressionEvaluator
                     Outcome = evaluatedExpression
                 });
             }),
+            // Expression: {Architecture}
```
brdeyo, line 213: Naming: CpuArchitecture.

This is the general name we use throughout the rest of our codebase.

src/VirtualClient/VirtualClient.Main/ExecuteProfileCommand.cs
```diff
@@ -825,9 +826,113 @@ private void InitializeProfile(ExecutionProfile profile)
 
             ValidationResult result = ExecutionProfileValidation.Instance.Validate(profile);
             result.ThrowIfInvalid();
+
+            // Process conditional parameters (ParametersOn feature)
+            await this.ProcessParametersOnAsync(profile, dependencies);
```
brdeyo, line 831: Similarly to how there is an extension method ExecutionProfile.Inline(), the logic to determine which parameters should be used from the conditional set should be an extension method as well. You may want to just extend the Inline() method to take in the IServiceCollection.

Recommendation:
I know that there are dependencies in IServiceCollection that cannot be referenced from the VirtualClient.Contracts project. We can likely just move the extensions method class to the VirtualClient.Core project to address the scenario.

brdeyo (changes requested):

PR #528, opened 2025-06-13
Filter Child Components dictated by the User-Provided Scenario Options.

cjhillbrand:
Filter the child components so that components that have the appropriate scenario(s) are ran.

### DIFF
--- a/src/VirtualClient/VirtualClient.Contracts/ComponentFactory.cs
+++ b/src/VirtualClient/VirtualClient.Contracts/ComponentFactory.cs
@@ -30,12 +30,22 @@ namespace VirtualClient.Contracts
         /// (vs. terminal).
         /// </param>
         /// <param name="logToFile">True to instruct the application to log output to files on the file system.</param>
+        /// <param name="includedScenarios">
+        /// When evaluating child components evaluate whether or not the child component should be 
+        /// included dictated by the scenarios provided on the command line. These scenarios must be included.
+        /// </param>
+        /// <param name="excludedScenarios">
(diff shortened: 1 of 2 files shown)


src/VirtualClient/VirtualClient.Core.UnitTests/ProfileExecutorTests.cs
```diff
@@ -52,8 +52,20 @@ public void SetupDefaults()
...
+                        components: new List<ExecutionProfileElement>()
+                        {
+                            new ExecutionProfileElement(
+                                "VirtualClient.TestExecutor",
+                                new Dictionary<string, IConvertible> { ["Scenario"] = "Scenario1" }),
+                            new ExecutionProfileElement(
+                                "VirtualClient.TestExecutor",
+                                new Dictionary<string, IConvertible> { ["Scenario"] = "Scenario2" }),
+                            new ExecutionProfileElement(
+                                "VirtualClient.TestExecutor",
```
yangpanMS, line 66: I see you setup the test correctly but I'm just a bit worried that all scenarios have duplicates (1-3), can you add another test specifically for nested scenario with Scenario4

yangpanMS (approved):

src/VirtualClient/VirtualClient.Core.UnitTests/ProfileExecutorTests.cs
cjhillbrand: Added a few more unit tests, and added a "Scenario4" component.

nchapagain001 (approved):

PR #531, opened 2025-06-18
Bug fix while working on running standalone VC without experiment ID …

nmalkapuram:
…explicitly passing in client server scenario


src/VirtualClient/VirtualClient.Actions/HammerDB/HammerDBExecutor.cs
```diff
@@ -84,7 +84,7 @@ public string SuperUserPassword
         {
             get
             {
-                byte[] hashBytes = SHA256.HashData(Encoding.UTF8.GetBytes(this.ExperimentId));
```
yangpanMS, line 87: standalone VC also have experimentID, and we use hammerDB or sysbench on workloads other than postgres like mysql

nmalkapuram: Thanks Yang for the comment. But by design VC's experimentID is not a mandatory parameter, why do we need to make it mandatory for these workloads when we can have it as optional parameter?

yangpanMS: it's not mandatory, we just provide a default which is random guid

yangpanMS: it always have an id so that you can find it in kusto even if you don't provide one

nmalkapuram: The issue here is when we run postgresql in client server scenario on two different machines manually, we would get two different experiment id's at client and server and the hashcode generated at client doesn't match with server, so client will fail in authenticating with server for the connections and connections would fail. This issue doesn't happen from juno because same exp id gets passed down to client and server machine.  This issue happens only when we run manually on two different machines. As we don't want exp id to be mandatory parameter, I changed it to common variable from which identical password is generated at both client and server.

yangpanMS: that is by design if you want the password to be different each time. Now if you don't want randomness we could just do a hash of "default"?

yangpanMS (approved):

PR #535, opened 2025-06-19
Allow for Optional Virtual Directory When Uploading Blob to maintain Local Folder Structure.

cjhillbrand:


brdeyo (changes requested):
I spoke with CJ on this. This is not quite the right solution to the problem. The implementation in another component "ToolkitExecutor" is where the root of the issue lies. We will fix this problem in that logic.

PR #536, opened 2025-06-20
Modifications in the Script based Executors

imadityaa:
Add alternate metrics file format for script executor and follow source directory structure when it moves logs

### DIFF
--- a/src/VirtualClient/VirtualClient.Actions/ScriptExecutor/JsonMetricsParser.cs
+++ b/src/VirtualClient/VirtualClient.Actions/ScriptExecutor/JsonMetricsParser.cs
@@ -5,9 +5,7 @@ namespace VirtualClient.Actions
 {
     using System;
     using System.Collections.Generic;
-    using System.ComponentModel;
     using System.Linq;
-    using System.Text.RegularExpressions;
     using global::VirtualClient;
     using global::VirtualClient.Actions;
     using global::VirtualClient.Contracts;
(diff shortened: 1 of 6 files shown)


src/VirtualClient/VirtualClient.Actions/ScriptExecutor/JsonMetricsParser.cs
```diff
@@ -46,20 +44,82 @@ public override IList<Metric> Parse()
...
+                                }
+                            }
+                        ]
+                    */
+                    foreach (var metricObj in token.Children<JObject>())
                     {
-                        throw new WorkloadResultsException(
-                            $"Invalid JSON metrics content formatting. The metric value for '{keyValuePair.Name}' is not a valid numeric data type.",
-                            ErrorReason.InvalidResults);
+                        string metricName = metricObj.Value<string>("metricName");
```
yangpanMS: If we are doing list of Metric, could we use JsonSerialization of Metric[] ? so you just use JsonConvert instead of custom logic

imadityaa: Modified

src/VirtualClient/VirtualClient.Actions/ScriptExecutor/JsonMetricsParser.cs
```diff
@@ -68,13 +137,29 @@ public override IList<Metric> Parse()
...
+                    $"Invalid JSON metrics content formatting. The metrics content must be in a valid JSON key/value pair format " +
+                    $"(e.g. {{ \"metric1\": 1234, \"metric2\": 987.65, \"metric3\": 32.0023481 }} ) or an array of Json objects.",
                     exc,
                     ErrorReason.InvalidResults);
             }
 
             return metrics;
         }
+
+        internal class CustomMetric
```
yangpanMS: Wait why we need this new class, can we use the main Metric object? The point is to standardize serialization and deserialization on the same object

imadityaa: The challenge is with the MetaData field, which is of type IDictionary<string, IConvertible>. Newtonsoft won't be able to deserialize directly to IConvertible type at runtime, so will might need a custom deserialize logic/ converter, which would be similar to my previous approach of custom logic

yangpanMS: try using this on Metadata, we have this on other classes
[JsonConverter(typeof(ParameterDictionaryJsonConverter))]

imadityaa: Thanks for this suggestion. I have modified the code.

yangpanMS (approved):

PR #543, opened 2025-07-02
Add option for MinimumIterations in ParallelLoopExecution

imadityaa:


src/VirtualClient/VirtualClient.Contracts/ParallelLoopExecution.cs
```diff
@@ -72,10 +83,17 @@ protected override Task ExecuteAsync(EventContext telemetryContext, Cancellation
         /// </summary>
         private async Task ExecuteComponentLoopAsync(VirtualClientComponent component, EventContext telemetryContext, CancellationToken cancellationToken)
         {
+            int iterationCount = 0;
             while (!cancellationToken.IsCancellationRequested)
             {
                 try
                 {
+                    if (this.timeoutTask.IsCompleted && iterationCount >= this.MinimumIteration)
```
yangpanMS: 1. I believe we want the sooner of the two
2. you could put this check at the end with the other exit checks

src/VirtualClient/VirtualClient.Contracts/ParallelLoopExecution.cs
```diff
@@ -84,14 +102,19 @@ private async Task ExecuteComponentLoopAsync(VirtualClientComponent component, E
 
                     // Execute the component task with timeout handling.
                     Task componentExecutionTask = component.ExecuteAsync(cancellationToken);
+
                     Task completedTask = await Task.WhenAny(componentExecutionTask, this.timeoutTask);
 
-                    if (completedTask == this.timeoutTask)
+                    if (completedTask == this.timeoutTask && iterationCount >= this.MinimumIteration)
```
yangpanMS: I believe we want the component to exit either when all components finished X times, or time is up, so we want the sooner of the two

imadityaa: Here the intent was to run the WLs for minimum X times, even if timeout has passed. Consider the case where we need the logs, then workload should run at least once, even if the timeout has crossed (Context CRC Packages chat).
If we exit when either components finished X times or timeout reached, then it would become 'MaximumIterations' instead of Minimum, which is intended.

src/VirtualClient/VirtualClient.Contracts/ParallelLoopExecution.cs
imadityaa: Added this check again here, for cases where timeout crossed before minimum iterations. Then to avoid any extra iteration, this check needs to happen initially.

The later check would also come handy in case where timeout happened after minimum iterations. Then, we don't want to wait for the on-going run of workloads, and rather stop then and there.

src/VirtualClient/VirtualClient.Contracts/ParallelLoopExecution.cs
imadityaa: By default, the minimum Iterations is set as 0. So, by default the timeout would be adhered to. This will change only when minimum iterations is not achieved before timeout

yangpanMS: I understand, but now we are creating a situation where user don't control the max time out, because it's looking for larger of the two. Tagging @brdeyo to decide

src/VirtualClient/VirtualClient.Contracts/ParallelLoopExecution.cs
```diff
@@ -42,6 +42,17 @@ public TimeSpan Duration
             }
         }
 
+        /// <summary>
+        /// The minimum number of times each child component should run. Default set to 0.
+        /// </summary>
+        public int MinimumIteration
```
brdeyo: Naming: "MinimumIterations" ....plural

brdeyo (dismissed):

src/VirtualClient/VirtualClient.Contracts/ParallelLoopExecution.cs
imadityaa: renamed it.

imadityaa: I'll have also added a documentation for this and other Custom Execution Orders like Parallel and Loop/Sequential in the PR for Loop/Sequential execution

yangpanMS (approved):

PR #546, opened 2025-07-07
Adding win-x64 support for nvidia-smi

deep1712:


src/VirtualClient/VirtualClient.Monitors/Nvidia-Smi/NvidiaSmiMonitor.cs
```diff
@@ -18,7 +18,7 @@ namespace VirtualClient.Monitors
     /// <summary>
     /// The Performance Counter Monitor for Virtual Client
     /// </summary>
-    [SupportedPlatforms("linux-arm64,linux-x64")]
+    [SupportedPlatforms("linux-arm64,linux-x64,win-x64")]
```
imadityaa, line 21: In Line 42, it checks for platform and only runs on Linux. We might need changes in ExecuteAsync for adding windows support

src/VirtualClient/VirtualClient.Monitors/Nvidia-Smi/NvidiaSmiMonitor.cs
```diff
@@ -39,34 +39,31 @@ protected override Task ExecuteAsync(EventContext telemetryContext, Cancellation
             {
                 try
                 {
-                    if (this.Platform == PlatformID.Unix)
+                    // Check that nvidia-smi is installed. If not, we exit the monitor.
+                    bool toolsetInstalled = await this.VerifyToolsetInstalledAsync(telemetryContext, cancellationToken);
```
yangpanMS, line 43: Do they use exactly the same commandline and output? also no need append .exe on windows? in some scripting environment you need explicitly .exe appended.

deep1712: It does not need any environment variable setting or .exe as suffix.
I works after installation of windows nvidia drivers.

psingla1210 (approved):

PR #550, opened 2025-07-17
Draft" TLS Support to OpenSSL workload - users/vdeepar/tls update

dheeparaj:


src/VirtualClient/VirtualClient.Main/profiles/Compete-OPENSSL-TLS.json
```diff
@@ -0,0 +1,102 @@
+{
+    "Description": "OpenSSL 3.5.0 Benchmark with TLS",
```
saibulusu, line 2: Should rename this profile to match the convention, PERF-CPU-OPENSSL-TLS.json for example.

src/VirtualClient/VirtualClient.Main/profiles/Compete-OPENSSL-TLS.json
```diff
@@ -0,0 +1,102 @@
...
+    "Metadata": {
+        "RecommendedMinimumExecutionTime": "01:00:00",
+        "SupportedPlatforms": "linux-x64,linux-arm64",
+        "SupportedOperatingSystems": "Ubuntu",
+        "logToFile":  true
+
+    },
+    "Parameters": {
+        "ServerIP": "10.1.0.92",
```
saibulusu: Does the IP address always have to be supplied by the user on the server-side for this profile?

saibulusu (commented):
Could you possibly add to the documentation for this so it's easier to understand what metrics to expect and what this profile does.

src/VirtualClient/VirtualClient.Main/profiles/Compete-OPENSSL-TLS.json
dheeparaj: we will use this naming convention until it is standardized and well accepted across teams. till then, this naming convention is recommended - (per Alex)

src/VirtualClient/VirtualClient.Main/profiles/Compete-OPENSSL-TLS.json
dheeparaj: IP address should not be there, it will get picked up from layout file, this profile is updated and pushed updated version

saibulusu (approved):

saibulusu (approved):

PR #554, opened 2025-07-24
Allow Virtual Client to use a certificate to authenticate to the proxy API

kmjjc:


src/VirtualClient/VirtualClient.Main/CommandBase.cs
```diff
@@ -588,11 +593,11 @@ protected virtual IServiceCollection InitializeDependencies(string[] args)
 
                 CommandBase.proxyApiDebugLoggers.Add(debugLogger);
 
-                blobStores.Add(DependencyFactory.CreateProxyBlobManager(new DependencyProxyStore(DependencyBlobStore.Content, this.ProxyApiUri), contentSource?.ToString(), debugLogger));
-                blobStores.Add(DependencyFactory.CreateProxyBlobManager(new DependencyProxyStore(DependencyBlobStore.Packages, this.ProxyApiUri), packageSource?.ToString(), debugLogger));
+                blobStores.Add(DependencyFactory.CreateProxyBlobManager(new DependencyProxyStore(DependencyBlobStore.Content, this.ProxyApiUri), contentSource?.ToString(), debugLogger, certificate));
```
kmjjc, line 597: we don't want the proxy URI here to include the certificate query parameters when setting the proxy API URI, change pending

src/VirtualClient/VirtualClient.Common/Rest/RestClientBuilder.cs
```diff
@@ -26,6 +31,7 @@ public RestClientBuilder(TimeSpan? timeout = null)
         {
             this.restClient = new RestClient();
```
cjhillbrand: Since we re-create this in the "build" method do we need to have a private field 'restClient'

src/VirtualClient/VirtualClient.Main/CommandBase.cs
```diff
@@ -618,6 +624,12 @@ protected virtual IServiceCollection InitializeDependencies(string[] args)
 
                     CommandBase.proxyApiDebugLoggers.Add(debugLogger);
 
+                    X509Certificate2 certificate = null;
```
cjhillbrand: Do we need to pass this certificate to any other method?

kmjjc: Good catch, on it

src/VirtualClient/VirtualClient.Common/Rest/RestClientBuilder.cs
kmjjc: Oh good catch. Actually we shoulnd't be recreating the restClient in the build method, the restClient headers are set in AddAuthorizationHeader and AddAcceptedMediaType and we don't want to overwrite them

cjhillbrand (approved):
looks good.

PR #555, opened 2025-07-25
prime95 fix

saibulusu:
Error message: "Process must exit before requested information can be determined."

original: 22d56276-2f46-47f6-b3e5-c2469e9e566e
wait for exit: 6729cb33-92fd-43ce-bd12-81e0b2bb1788
5 seconds: 5e8cd34b-9518-427f-8ae0-05047148ba32


src/VirtualClient/VirtualClient.Actions/Prime95/Prime95Executor.cs
```diff
@@ -291,6 +291,8 @@ await this.Logger.LogMessageAsync($"{this.TypeName}.ExecuteWorkload", telemetryC
                         {
                             await this.WaitAsync(explicitTimeout, cancellationToken);
                             process.SafeKill();
+                            // await process.WaitForExitAsync(cancellationToken);
```
yangpanMS: Make sure this doesn't add delay to unit tests, if it does, find a way to mock it

yangpanMS (approved):

yangpanMS:
Does this work for you? if not check my changes in #559

saibulusu:
> Does this work for you? if not check my changes in #559

It still doesn't work when I use the process manager kill method, it produces the 137 OOM error when I use a VM of size Standard D64ds v6 (64 vcpus, 256 GiB memory).

PR #557, opened 2025-07-30
fio profiles fixes

saibulusu:
1. Moving linux package installation prior to disk initialization.
2. Removing libaio-dev(el) as it seems to be not needed.

### DIFF
--- a/src/VirtualClient/VirtualClient.Actions.FunctionalTests/FioProfileTests.cs
+++ b/src/VirtualClient/VirtualClient.Actions.FunctionalTests/FioProfileTests.cs
@@ -79,7 +79,6 @@ namespace VirtualClient.Actions
 
                 // Apt packages expectations
                 // There are a few Apt packages that must be installed for the FIO workload to run.
-                WorkloadAssert.AptPackageInstalled(this.mockFixture, "libaio-dev");
                 WorkloadAssert.AptPackageInstalled(this.mockFixture, "fio");
             }
         }


saibulusu:
Is there a reason why we had to move format disks prior to linux package installation? It seems lshw is required on azlinux prior to disk initialization.

src/VirtualClient/VirtualClient.Main/profiles/PERF-IO-FIO-DISCOVERY.json
```diff
@@ -527,16 +537,6 @@
                 "DiskFilter": "$.Parameters.DiskFilter"
             }
         },
-        {
-            "Type": "LinuxPackageInstallation",
-            "Parameters": {
-                "Scenario": "InstallLinuxPackages",
```
yangpanMS, line 533: which os did you test on? libaio now comes in all distro we use including ubuntu24 and azlinux3?

yangpanMS (approved):
Approved but make sure the package you removed are now available on all major OS we test in

src/VirtualClient/VirtualClient.Main/profiles/PERF-IO-FIO-DISCOVERY.json
saibulusu: Ubuntu 24, Ubuntu 22, AzLinux 3. I removed the package because it doesn't seem necessary to run fio.
libaio-dev(el) is not installed with fio on AzLinux 3, but it does install with fio on both Ubuntu vesions.

PR #562, opened 2025-08-10
fio discovery profile expansion

saibulusu:
**Purpose of expansion**: To eventually make the following Compete profiles obsolete and use parameters with a profile in the public repo.
- [PERF-FIO-MAXIMA-SWEEP-AzLinux.json](https://msazure.visualstudio.com/One/_git/CRC-ACC?path=/src3/Compete/Profiles/InProgress/PERF-FIO-MAXIMA-SWEEP-AzLinux.json&version=GBmaster)
- [PERF-FIO-MAXIMA-SWEEP-Ubuntu24-custombuild.json](https://msazure.visualstudio.com/One/_git/CRC-ACC?path=/src3/Compete/Profiles/InProgress/PERF-FIO-MAXIMA-SWEEP-Ubuntu24-custombuild.json&version=GBmaster)
- [PERF-FIO-MAXIMA-SWEEP-Ubuntu24.json](https://msazure.visualstudio.com/One/_git/CRC-ACC?path=/src3/Compete/Profiles/InProgress/PERF-FIO-MAXIMA-SWEEP-Ubuntu24.json&version=GBmaster)

**Changes made to FioDiscoveryExecutor**: Removed the nested for loop for max threads, io type, and block size. So now only queue depth is iterated over/permitted as a list parameter.
**Changes made to PERF-IO-FIO-DISCOVERY.json**: Added 1k block size.

### DIFF
--- a/src/VirtualClient/VirtualClient.Actions/FIO/oltp-c.fio.jobfile
+++ b/src/VirtualClient/VirtualClient.Actions/FIO/oltp-c.fio.jobfile
@@ -21,7 +21,7 @@ bs=${randomreadblocksize}
 iodepth=${randomreadiodepth}
 numjobs=${randomreadnumjobs}
 rate_iops=${randomreadiops}
-runtime=${durationsec}
+runtime=${duration}
 rw=randread
 
 # Represents writing of data in sql/OLTP database.
@@ -31,7 +31,7 @@ bs=${randomwriteblocksize}
 iodepth=${randomwriteiodepth}
 numjobs=${randomwritenumjobs}
 rate_iops=${randomwriteiops}
-runtime=${durationsec}
+runtime=${duration}
 rw=randwrite
 
 # Represents log readers of sql/OLTP database.
@@ -42,7 +42,7 @@ filename=${sequentialiofilepath}
 iodepth=${sequentialreadiodepth}
 numjobs=${sequentialreadnumjobs}
 rate_iops=${sequentialreadiops}
-runtime=${durationsec}
+runtime=${duration}
 rw=read
 size=${sequentialiofilesize}
 
@@ -54,7 +54,7 @@ filename=${sequentialiofilepath}
 iodepth=${sequentialwriteiodepth}
 numjobs=${sequentialwritenumjobs}
 rate_iops=${sequentialwriteiops}
-runtime=${durationsec}
+runtime=${duration}
 rw=write
 size=${sequentialiofilesize}
 
--- a/src/VirtualClient/VirtualClient.Actions/FIO/oltp-c2-operations.fio.jobfile
+++ b/src/VirtualClient/VirtualClient.Actions/FIO/oltp-c2-operations.fio.jobfile
@@ -18,7 +18,7 @@ bs=${randomreadblocksize}
 iodepth=${randomreadiodepth}
 numjobs=${randomreadnumjobs}
 rate_iops=2708
-runtime=${durationsec}
+runtime=${duration}
 rw=randread
 size=${randomiofilesize}
 
@@ -29,7 +29,7 @@ bs=${randomwriteblocksize}
 iodepth=${randomwriteiodepth}
 numjobs=${randomwritenumjobs}
 rate_iops=2128
-runtime=${durationsec}
+runtime=${duration}
 rw=randwrite
 size=${randomiofilesize}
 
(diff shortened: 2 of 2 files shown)


src/VirtualClient/VirtualClient.Main/profiles/PERF-IO-FIO-DISCOVERY-v2.json
```diff
@@ -0,0 +1,93 @@
...
+                "DiskFillSize": "$.Parameters.DiskFillSize",
+                "DiskFill": true,
+                "DeleteTestFilesOnFinish": false,
+                "DiskFilter": "$.Parameters.DiskFilter"
+            }
+        },
+        {
+            "Type": "FioDiscoveryExecutor",
+            "Parameters": {
+                "Scenario": "FIOAction",
```
brdeyo: The scenario name is important. The original profile had a discovery executor per block size to ensure that the Scenario was clear and functional. The Scenario for example can be used to run a specific Action in the set.

e.g.
VirtualClient.exe --profile=PERF-IO-FIO-DISCOVERY.json --scenarios=RandomRead_4k_BlockSize,RandomRead_8k_BlockSize

With this profile, this functional feature will no longer work as intended.

brdeyo: I understand the problem you are trying to solve, but I would recommend solving it a different way. You can for example modify/correct the metadata being supplied to the Metrics so that the "metadata_metrics" metadata set is correct. This looks like a code bug more than an issue with how the profile is currently defined.

brdeyo (changes requested):

src/VirtualClient/VirtualClient.Main/profiles/PERF-IO-FIO-DISCOVERY-v2.json
saibulusu: Should we be iterating over multiple block sizes and io types with this executor then?

saibulusu: Also we could use the block size and io type parameters to run different combinations.
e.g.
VirtualClient.exe --profile=PERF-IO-FIO-DISCOVERY-v2.json --parameters="IOType=randread,randwrite;bs=1k,4k"

But this would run all combinations with those given parameters.

saibulusu: Removed the v2 profile, and added a block size to PERF-IO-FIO-DISCOVERY.json.

src/VirtualClient/VirtualClient.Main/profiles/PERF-IO-FIO-DISCOVERY.json
```diff
@@ -31,6 +30,26 @@
...
         },
+        {
+            "Type": "FioDiscoveryExecutor",
+            "Parameters": {
+                "Scenario": "RandomRead_1k_BlockSize",
+                "BlockSize": "1k",
+                "CommandLine": "--ioengine={Engine} --size={FileSize} --rw={IOType} --bs={BlockSize} --direct={DirectIO} --ramp_time=15 --runtime={DurationSec} --time_based --overwrite=1 --thread --group_reporting --output-format=json",
+                "Engine": "$.Parameters.Engine",
+                "FileSize": "$.Parameters.FileSize",
+                "DurationSec": "180",
```
brdeyo: Please do me a favor and change 'DurationSec' to 'Duration' following the same pattern used in the PERF-IO-FIO.json profile. Add a global parameter with a default value of 00:03:00 to match the 180 seconds or 3 mins to the top of the profile as well to be consistent with the base FIO profile.

brdeyo: Please do the same to both profiles in the changeset here. Thanks!!

saibulusu: For FioDiscoveryExecutor I removed the field from the executor completely. But for FioMultiThroughputExecutor I converted it within the executor to seconds as it is used in the executor.

brdeyo (approved):

PR #632, opened 2026-02-05
Adding more flexibility to networking workload

deep1712:


bistmadhav:
Review comment - The refactored command-line generation is not reached consistently. CPSExecutor.GetCommandLineArguments() still returns null, and LatteExecutor.GetCommandLineArguments() returns an empty string. NTttcp and SockPerf server are created through the component-copy constructors in CreateWorkloadExecutor(), but those constructors do not initialize the new CommandLine* properties. Consequently, several production paths will launch workloads with null or empty arguments. Command construction should occur inside GetCommandLineArguments() after InitializeAsync, not selectively inside constructors. Please also add tests asserting the exact generated command for every tool, role, and platform.

PR #686, opened 2026-04-11
Fix HammerDB/PostgreSQL: use hex password to avoid TCL parsing error

AlexW-F:
Base64 password contains + and = characters which break HammerDB 4.12's TCL diset command argument parsing. Changed to hex encoding which produces alphanumeric-only output across all three components that share the SuperUserPassword:
- HammerDBExecutor
- PostgreSQLServerConfiguration
- SysbenchExecutor

All three must use the same encoding since they share the same PostgreSQL credential.

This has been the sole cause of HammerDB parse failures since Mar 16 (~1700 errors/day).


src/VirtualClient/VirtualClient.Actions/HammerDB/HammerDBExecutor.cs
```diff
@@ -98,7 +98,7 @@ public string SuperUserPassword
             get
             {
                 byte[] hashBytes = SHA256.HashData(Encoding.UTF8.GetBytes("default"));
-                return Convert.ToBase64String(hashBytes);
+                return Convert.ToHexString(hashBytes);
```
nchapagain001, line 101: Add UT next please.

nchapagain001:
Don't forget to up the version.

nchapagain001 (approved):

nchapagain001 (approved):

PR #692, opened 2026-04-16
Add NTTTCP full duplex mode

rudraptpsingh:
Enable bidirectional network testing where both nodes run sender and receiver simultaneously, producing separate TX/RX throughput metrics per direction.

Core changes:
- NTttcpExecutor: Add DuplexMode property, ExecuteFullDuplexWorkloadAsync with parallel sender+receiver processes, CaptureDirectionalMetrics for per-direction telemetry, role-aware port assignment (forward/reverse), direction-aware -m IP mapping for correct sender→receiver targeting
- NetworkingWorkloadState: Add DuplexMode property for client-server state sync
- NetworkingWorkloadExecutor: Pass DuplexMode through orchestration instructions

Profile:
- PERF-NETWORK-NTTTCP-FULLDUPLEX.json: 6 TCP scenarios (T1/T32 x 4K/64K/256K) with DuplexMode=Full, same dependencies as half-duplex profile

Tests (21 tests):
- NTttcpFullDuplexTests: IsFullDuplex property, ReversePort offset, command-line flag/port/buffer verification for both platforms, execution tests for client send+receive on Linux/Windows, results file separation, half-duplex backward compatibility, NetworkingWorkloadState serialization with DuplexMode

Upstream fixes (pre-existing bugs in main):
- ProcessProxy.cs: Fix catch clause ordering (general catch before specific)
- CommandBase.cs: Fix NullReferenceException on isolationTargets.TargetPackages

Verified on real hardware: win-x64 <-> win-arm64
TCP 4K T1: Send 3226 Mbps / Receive 1455 Mbps
TCP 64K T1: Send 6490 Mbps / Receive 6163 Mbps

### DIFF
--- a/src/VirtualClient/VirtualClient.Main/profiles/PERF-NETWORK-NTTTCP.json
+++ b/src/VirtualClient/VirtualClient.Main/profiles/PERF-NETWORK-NTTTCP.json
@@ -11,7 +11,8 @@
         "NTttcpDuration": "00:01:00",
         "ProfilingEnabled": false,
         "ProfilingMode": "None",
-        "TestDuration": "00:01:00"
+        "TestDuration": "00:01:00",
+        "DuplexMode": "Half"
     },
     "Actions": [
         {
@@ -28,7 +29,8 @@
                 "Port": "$.Parameters.NTttcpPort",
                 "ProfilingScenario": "NTttcp_TCP_4K_Buffer_T1",
                 "ProfilingEnabled": "$.Parameters.ProfilingEnabled",
-                "ProfilingMode": "$.Parameters.ProfilingMode"
+                "ProfilingMode": "$.Parameters.ProfilingMode",
+                "DuplexMode": "$.Parameters.DuplexMode"
             }
         },
         {
@@ -45,7 +47,8 @@
                 "Port": "$.Parameters.NTttcpPort",
                 "ProfilingScenario": "NTttcp_TCP_64K_Buffer_T1",
                 "ProfilingEnabled": "$.Parameters.ProfilingEnabled",
-                "ProfilingMode": "$.Parameters.ProfilingMode"
+                "ProfilingMode": "$.Parameters.ProfilingMode",
+                "DuplexMode": "$.Parameters.DuplexMode"
             }
         },
         {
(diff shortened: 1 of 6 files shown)


brdeyo (changes requested):
Please hold up on merging these changes until we have completed the 3.0 production release of VC. We are aiming at having that out today Thursday April 16th. However, I want to prevent anymore major changes to VC core or workloads until we finalize.

src/VirtualClient/VirtualClient.Common/ProcessProxy.cs
```diff
@@ -358,6 +358,10 @@ public virtual async Task WaitForExitAsync(CancellationToken cancellationToken,
 
                 await this.WaitForExitCodeConfirmationAsync(cancellationToken, timeout);
             }
+            catch (TimeoutException) when (timeout != null)
```
brdeyo: Not needed. There is a generic "catch" block. This catches ALL exceptions including TimeoutException.

brdeyo (changes requested):

brdeyo (approved):

src/VirtualClient/VirtualClient.Main/profiles/PERF-NETWORK-NTTTCP.json
```diff
@@ -11,7 +11,8 @@
         "NTttcpDuration": "00:01:00",
         "ProfilingEnabled": false,
         "ProfilingMode": "None",
-        "TestDuration": "00:01:00"
+        "TestDuration": "00:01:00",
+        "DuplexMode": "Half"
```
nchapagain001, line 15: Please confirm my understanding:
Was DuplexMode: Half original behavior?

nchapagain001 (approved):
Discussed offline.

rudraptpsingh:
Update: Fixed full-duplex race condition

Added a cross-machine handshake to fix a timing issue in full-duplex mode where the server's sender would start connecting to the client's reverse port before the client's receiver was listening.

Root cause: After the server sets state=Running (via ConfirmProcessRunningAsync), it starts its sender 2 seconds later. But the client only detects Running after the 10-second polling interval, then starts its receiver. By then, the server's sender has been retrying for ~8+ seconds. Linux ntttcp's ~11-second retry budget gets exhausted → "Unexpected disconnect".

Fix: Uses the existing VirtualClient state API to add a handshake step:

Client starts its receiver → signals FullDuplexClientReceiverReady to server
Server polls for this signal → only then starts its sender
VM test results (Win=Client, Linux=Server):

Metric	Before	After
Success rate	14% (1/7)	100% (5/5)
Errors	6 "Unexpected disconnect"	0
Win→Linux	907 Mbps	1,208 Mbps
Linux→Win	3,588 Mbps	3,075 Mbps

imadityaa (approved):

PR #714, opened 2026-05-22
Bump DiskSpd to v2.2.0 and fix metrics parser for new CPU table format

rudraptpsingh:
- Promote BlobName to Parameters.PackageBlobName (default: diskspd.2.2.0.zip)
- Allows runtime override without profile edit
- diskspd.2.2.0.zip uploaded packages

### DIFF
--- a/src/VirtualClient/VirtualClient.Actions.UnitTests/DiskSpd/DiskSpdMetricsParserTests.cs
+++ b/src/VirtualClient/VirtualClient.Actions.UnitTests/DiskSpd/DiskSpdMetricsParserTests.cs
@@ -200,6 +200,43 @@ namespace VirtualClient.Actions
         }
 
         [Test]
+        public void DiskSpdParserVerifyWriteOnlyV220FormatWithSocketNodeCoreColumns()
+        {
+            // DiskSpd v2.2.0 changed the CPU table to include Socket | Node | Group | Core | CPU columns.
+            // The parser must normalise this extended header without throwing a KeyNotFoundException.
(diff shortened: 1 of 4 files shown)


brdeyo (changes requested):
We cannot do this as it is in conflict with an important aspect of VC ensuring users cannot make changes to our out-of-box profiles that have the potential to cause breaks or to deviate the core purpose of the profile.

Core workload packages are rarely something we allow users to override in parameters. I would not like to change this profile this way in open source. Some options you may consider are to use a custom profile for your specific purpose or to simply set the package version to 2.2.0 in the DependencyPackageInstallation. It is best it is "set in stone".

rudraptpsingh:
> We cannot do this as it is in conflict with an important aspect of VC ensuring users cannot make changes to our out-of-box profiles that have the potential to cause breaks or to deviate the core purpose of the profile.
> 
> Core workload packages are rarely something we allow users to override in parameters. I would not like to change this profile this way in open source. Some options you may consider are to use a custom profile for your specific purpose or to simply set the package version to 2.2.0 in the DependencyPackageInstallation. It is best it is "set in stone".

Updated to update the package version only in profile. Also fixes DiskSpdMetricsParser to handle the v2.2.0 CPU table format change — DiskSpd v2.2.0 added Socket | Node | Core columns to the CPU section header, which caused the parser's Preprocess() to key the section as "Socket | Node | CPU" instead of "CPU", resulting in a KeyNotFoundException at parse time. The fix detects the extended header and normalizes it to the legacy Group | CPU format before sectionizing. Includes a new unit test and example output file based on a real v2.2.0 run on a 160-core Windows host. All existing DiskSpd parser tests pass.

brdeyo (approved):

PR #719, opened 2026-05-27
Add DuplexMode parameter to PERF-NETWORK profile and bump version to …

rudraptpsingh:
…3.2.10

Propagates the NTTTCP full-duplex change (#692) to the PERF-NETWORK suite profile, which was missed in the original PR. Adds DuplexMode (default Half) to the Parameters block and threads it through all 11 NTttcp actions (9 TCP + 2 UDP).


brdeyo (changes requested):
This is a core profile used in data center sustainment. We cannot make a major change like this without affecting the ability to weigh current NTttcp performance numbers against past numbers.

Set the "DuplexMode" parameter to "Half"...the previous default behavior. For the handful of cases where there is a wish for full duplex, this option should be set on the command line using --parameters="NTttcpDuplexMode=Full" (see parameter rename comment below).

Additionally, this profile runs 4 different toolsets. In order to avoid confusion, keep to the same naming conventions as we have for other per-tool settings. In this case change the global parameter from "DuplexMode" to "NTttcpDuplexMode". Make this same change to the PERF-NETWORK-NTTTCP.json profile so we are consistent in name and value across both profiles for global parameters.

You can leave the parameter named "DuplexMode" for the NetworkWorkloadExecutor however.

brdeyo:
> …3.2.10
> 
> Propagates the NTTTCP full-duplex change (#692) to the PERF-NETWORK suite profile, which was missed in the original PR. Adds DuplexMode (default Half) to the Parameters block and threads it through all 11 NTttcp actions (9 TCP + 2 UDP).

We cannot make blanket statements like this without significant evidence and need to be careful taking actions without evidence. We have been running this profile with NTttcp this way for 4 - 5 years in the data center across 10,000s of systems with success (including on ARM systems). The configuration was provided by the Azure Networking team themselves and is thus not a random set of guesses.

rudraptpsingh:
> > …3.2.10
> > Propagates the NTTTCP full-duplex change (#692) to the PERF-NETWORK suite profile, which was missed in the original PR. Adds DuplexMode (default Half) to the Parameters block and threads it through all 11 NTttcp actions (9 TCP + 2 UDP).
> 
> We cannot make blanket statements like this without significant evidence and need to be careful taking actions without evidence. We have been running this profile with NTttcp this way for 4 - 5 years in the data center across 10,000s of systems with success (including on ARM systems). The configuration was provided by the Azure Networking team themselves and is thus not a random set of guesses.

I don't think I fully understand the issue in the statement above. By original PR I meant the NTTTCP full duplex mode PR. I am not diminishing the NTTTCP execution results till now, but duplex mode definitely seems an improvement over half duplex that we have currently.

I agree we should have it half duplex mode by default though.

brdeyo (approved):

brdeyo (approved):

nmalkapuram (approved):

brdeyo (changes requested):
Need to update the VERSION

brdeyo (approved):

PR #724, opened 2026-06-02
mongodb migration to open-source

saibulusu:
Ran on a small db size (500000 records) in client/server with just workload a - aa4d57d9-3e61-4f3e-9813-f7bf79c88445


src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-SINGLE-CLIENTSERVER.json
```diff
@@ -0,0 +1,252 @@
...
+                "JdkPackageName": "$.Parameters.JdkPackageName",
+                "RecordCount": "$.Parameters.RecordCount",
+                "ThreadCount": "$.Parameters.ThreadCount",
+                "WorkloadName": "$.Parameters.WorkloadName"
+            }
+        },
+        {
+            "Type": "MongoDBClientExecutor",
+            "Parameters": {
+                "Scenario": "{WorkloadName}_fieldcount{FieldCount}_fieldlength{FieldLength}_OpCount{OperationCount}",
```
brdeyo: Scenarios must be string literals, no parameter references. Add in a parameter "MetricScenario" for the current implementation and update the code to use that for metrics if it is referencing the "Scenario" to use the other parameter instead. Additionally, the names should be consistent in casing...not mostly lower but then some upper/pascal (e.g. OpCount).

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-SINGLE-CLIENTSERVER.json
```diff
@@ -0,0 +1,252 @@
...
+        "WorkloadName": "workloada",
+        "Duration": "00:05:00",
+        "Database": "mongodb",
+        "Port": "27017",
+        "FieldCount": "128",
+        "FieldLength": "128",
+        "OperationCount": "5000000",
+        "RecordCount": "500000",
+        "YCSBPackageName": "ycsb",
+        "JdkPackageName": "javadevelopmentkit",
```
brdeyo: Package names should not be global parameters. Move these names to the usage locations directly:
- YCSBPackageName
- JdkPackageName

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-SINGLE-CLIENTSERVER.json
```diff
@@ -0,0 +1,252 @@
...
+        "DiskFilter": "BiggestSize",
+        "dbSize": "small",
+        "CommentOnDbSize": "dbSize is approx 8-10 GB"
```
brdeyo: Rename "CommentOnDbSize" to "Comments". CommentOnDbSize is being used to define notes/metadata. It is NOT functional and would thus not typically be defined as a parameter. I presume this is for telemetry clarification at some point. However, no need to be so explicit in the name.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-SINGLE-CLIENTSERVER.json
```diff
@@ -0,0 +1,252 @@
...
+            "FieldCount": "128",
+            "FieldLength": "128",
+            "CommentOnDbSize": "dbSize is approx 8-10 GB"
+        },
+        {
+            "Condition": "{calculate(\"{dbSize}\" == \"medium\")}",
+            "RecordCount": "2500000",
+            "FieldCount": "128",
+            "FieldLength": "128",
+            "Duration": "00:30:00",
```
brdeyo: Duration should not be shifting around based on the DB size. This creates a "black box" scenario for the user. To be consistent with ALL of the rest of our use cases where we allow a duration to be defined, the value set in the global parameters should be used and not changed in the ParametersOn. Remove it from the conditions in that section.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-SINGLE-CLIENTSERVER.json
```diff
@@ -0,0 +1,252 @@
...
+        "SupportedPlatforms": "linux-x64,linux-arm64",
+        "SupportedOperatingSystems": "Ubuntu"
+    },
+    "Parameters": {
+        "ThreadCount": "{calculate({LogicalCoreCount}/2)}",
+        "WorkloadName": "workloada",
+        "Duration": "00:05:00",
+        "Database": "mongodb",
+        "Port": "27017",
+        "FieldCount": "128",
```
brdeyo: FieldCount and FieldLength should not be defined here. Overriding these values will cause the purpose of the "DBSize" parameter to be skewed. Remove both from here.

Just add a default condition block to the ParametersOn section at the very end with the "Condition" set to true. This will set the defaults.

brdeyo: Same for "RecordCount". If we are going to use a parameter "DBSize" that dictates the desired size of the database by mapping different DB size values to FieldCount, FieldLength and RecordCount values then none of those values should be in the global parameters such that the user could do this on the command line.

--parameters="DBSize=small,,,RecordCount = 100000000000000000000000000000000".

As you can ascertain, that number of records would definitely NOT result in a small database. Furthermore however, that value would not be used because the ParametersOn section would have a conditional block to override it. Now the user is confused.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-SINGLE-CLIENTSERVER.json
```diff
@@ -0,0 +1,252 @@
...
+        "FieldLength": "128",
+        "OperationCount": "5000000",
```
brdeyo: OperationCount seems to be in direct conflict with Duration...i.e. I could set an operation count so large that it could not actually finish within the duration. Remove OperationCount from the global parameters and define the value explicitly in the places it is used.

brdeyo (changes requested):

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-SINGLE-CLIENTSERVER.json
saibulusu: Moved to usage locations.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-SINGLE-CLIENTSERVER.json
saibulusu: Added notes in metadata for this.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-SINGLE-CLIENTSERVER.json
saibulusu: Just using RecordCount, no longer using dbSize.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-SINGLE-CLIENTSERVER.json
saibulusu: Removed OperationCount as a global parameter.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-SINGLE-CLIENTSERVER.json
saibulusu: Duration is set as a global parameter. No more conditional parameters.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-SINGLE-CLIENTSERVER.json
saibulusu: Changed all scenario names.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-YCSB.json
```diff
@@ -0,0 +1,334 @@
...
+                "OperationCount": "5000000",
+                "YCSBPackageName": "ycsb",
+                "JdkPackageName": "javadevelopmentkit",
+                "RecordCount": "$.Parameters.RecordCount",
+                "FieldCount": "128",
+                "FieldLength": "128"
+            }
+        },
+        {
+            "Type": "MongoDBClientExecutor",
```
brdeyo, line 141: I missed this originally.

I cannot quite understand why we are dropping the database midstream but would want to NOT do that. To be consistent with our typical DB process, the profile execution should create the database once at the beginning of a run and then run the various scenarios against it. Dropping the database midway does not create an apples to apples situation and additionally sets up the possibility of a corrupted run if the step fails.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-YCSB.json
```diff
@@ -0,0 +1,334 @@
...
+                "Scenario": "serversetup",
+                "Role": "Server",
+                "Port": "$.Parameters.Port",
+                "Database": "$.Parameters.Database",
+                "DiskFilter": "$.Parameters.DiskFilter"
+            }
+        },
+        {
+            "Type": "MongoDBClientExecutor",
+            "Parameters": {
```
brdeyo, line 30: Realistically, this step should be a Dependency so that it is executed only once when VC runs. I would expect VC to support the following scenarios:

1) The database is dropped and recreated at the start of each distinct VC execution.
2) Given the user wants to create the database ONLY once across multiple distinct VC executions, they can do the following:
    step 1: VirtualClient.exe --profile=PERF-MONGODB-YCSB.json --dependencies -> to create the database
    step 2: VirtualClient.exe --profile=PERF-MONGODB-YCSB.json  --scenarios=-initialize_database -> and run 
                 this any number of times to use the DB created in step 1 over and over (without making any
                 modifications to the DB).

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-YCSB.json
```diff
@@ -0,0 +1,334 @@
...
+                "Scenario": "loaddatabase_initial",
```
brdeyo: Rename scenario from "loaddatabase_initial" to "initialize_database"

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-YCSB.json
```diff
@@ -0,0 +1,334 @@
...
+                "JdkPackageName": "javadevelopmentkit",
+                "RecordCount": "$.Parameters.RecordCount",
+                "FieldCount": "128",
+                "FieldLength": "128"
+            }
+        },
+        {
+            "Type": "MongoDBClientExecutor",
+            "Parameters": {
+                "Scenario": "read50_rmw50",
```
brdeyo: What is rmw? We should just spell it out to be consistent with the other naming conventions (e.g. readwrite50)

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-YCSB.json
```diff
@@ -0,0 +1,334 @@
...
+                "YCSBPackageName": "ycsb",
+                "JdkPackageName": "javadevelopmentkit",
+                "RecordCount": "$.Parameters.RecordCount",
+                "ThreadCount": "$.Parameters.ThreadCount"
+            }
+        },
+        {
+            "Type": "MongoDBClientExecutor",
+            "Parameters": {
+                "Scenario": "scan95_insert05",
```
brdeyo: Do a quick bit of research to determine if this is a real world example. Seems like we are just inventing a scenario here and that testing a "scan" only would be the better. Based on a quick bit of research myself, these are the recommended real-world usage cases that we can use to map here.

When conducting cloud customer benchmarking for MongoDB using the **Yahoo! Cloud Serving Benchmark (YCSB)**, engineers map tests to standard YCSB core workloads. These workloads mirror common real-world cloud application architectures, user traffic variations, and system usage patterns.

Below is a summary of **6 standard core YCSB benchmarks** mapped to real usage patterns, followed by **2 custom benchmarking scenarios** frequently utilized by cloud customers to replicate advanced production behaviors.

---

### Standard YCSB Core Benchmarks

#### 1. Workload A: Update-Heavy (50% Read / 50% Update) - Scenario = update_heavy, MetricScenario = update_heavy_50%_read_50%_update, MetricCategory=50000 records

* **Real Usage Pattern:** **E-commerce Session Stores & Shopping Carts.** This simulates an active user session where actions are constantly written or updated (e.g., adding an item to a cart, updating viewed items) while simultaneously reading the current state of the session.
* **Database Impact:** Stresses both the read path (WiredTiger cache hits) and the write path (journaling, disk I/O, and lock concurrency under simultaneous reads and updates).

#### 2. Workload B: Read-Mostly (95% Read / 5% Update) - Scenario = read_mostly, MetricScenario=read_mostly_50%_read_50%_update, MetricCategory=50000 records

* **Real Usage Pattern:** **Social Media & Photo Tagging.** In applications like a photo feed or tagging service, users spend the vast majority of their time viewing (reading) existing content, with occasional actions like adding a tag or liking a post (updating metadata).
* **Database Impact:** Tests the efficiency of MongoDB’s indexing and heavy read caching capabilities while validating that light write spikes do not introduce reading degradation.

#### 3. Workload C: Read-Only (100% Read) - Scenario = read_only, MetricScenario=read_only_100%_read, MetricCategory=50000 records

* **Real Usage Pattern:** **Content Delivery & User Profile Caching.** This maps directly to scenarios where a system serves relatively static data—such as read-only lookups for product catalogs, global user settings, or backend master data that is populated out-of-band by a separate analytics processor.
* **Database Impact:** Measures maximum clean read throughput (IOPS) and memory performance. Customers often use this to determine the optimal RAM size for their working set.

#### 4. Workload D: Read-Latest (95% Read / 5% Insert) - Scenario = read_latest, MetricScenario=read_latest_95%_read_5%_insert, MetricCategory=50000 records

* **Real Usage Pattern:** **Status Updates & Microblogging (e.g., Timeline Feeds).** This models a system where users are constantly publishing short status updates or tweets (inserts), and popular demand heavily targets the most recently published entries rather than archival data.
* **Database Impact:** Highlights cache recency behaviors. It stresses how fast newly inserted B-tree leaf nodes can be read back from memory across cloud node replicas.

#### 5. Workload E: Short Ranges (95% Scan / 5% Insert) - Scenario = short_range_scan, MetricScenario=short_range_scan_95%_scan_5%_insert, MetricCategory=50000 records

* **Real Usage Pattern:** **Threaded Messaging & Forums.** Instead of fetching a single record via a unique ID, the client queries ranges of documents (e.g., pulling the latest 20 comments clustered by a thread ID).
* **Database Impact:** Evaluates MongoDB's index cursor performance, horizontal range scanning, and pagination handling rather than point-in-time point lookups.

#### 6. Workload F: Read-Modify-Write (50% Read / 50% Read-Modify-Write) - Scenario = read_modify_write, MetricScenario=read_modify_write_50%_read_50%_read_modify_write, MetricCategory=50000 records

* **Real Usage Pattern:** **User Account Wallets & Gaming Leaderboards.** In this pattern, an application reads a user's record (e.g., existing bank balance or game score), calculates a modification client-side, and securely saves the updated document back to the database.
* **Database Impact:** Evaluates read-update lock behavior, document-level concurrency controls, and the latency added by multiple round-trips over the cloud network.

#### 7. The Peak Event Shift (0% Read / 100% Insert)  - Scenario = sudden_burst, MetricScenario=sudden_burst_0%_read_100%_insert, MetricCategory=50000 records

* **Real Usage Pattern:** **IoT Fleet Telemetry & Flash Sale Registration.** This replicates a sudden burst or continuous ingestion of raw data—such as thousands of connected vehicles dumping location logs every second or a sudden onslaught of black-friday ticketing requests.
* **Database Impact:** Forces MongoDB to flush its internal memory journal to persistent cloud block storage (like AWS EBS or Google Persistent Disk) continuously. This reveals how fast the cloud disk can sustain constant random/sequential write pressure.

---

### Documentation & Implementation References

To set up, configure, and execute these workloads on a MongoDB cluster, you can utilize the following official guides:

* **Official Framework Reference:** [[YCSB GitHub Core Workloads Specification](https://www.google.com/search?q=https://github.com/brianfrankcooper/YCSB/wiki/Core-Workloads)](https://www.google.com/search?q=https://github.com/brianfrankcooper/YCSB/wiki/Core-Workloads) — Detailed breakdown of properties, distributions (such as Zipfian), and core workloads configurations.
* **MongoDB Driver Compatibility:** [[MongoDB Labs YCSB Repository](https://github.com/mongodb-labs/YCSB)](https://github.com/mongodb-labs/YCSB) — The preferred fork maintained by MongoDB engineers to handle compatibility with modern versions of MongoDB (5.0, 6.0, and 7.0+) following legacy opcode migrations.
* **Cloud Infrastructure Implementation Examples:** * [[Arm Server Optimization & MongoDB YCSB Benchmarking Guide](https://learn.arm.com/learning-paths/servers-and-cloud-computing/glibc-with-lse/mongo_benchmark/)](https://learn.arm.com/learning-paths/servers-and-cloud-computing/glibc-with-lse/mongo_benchmark/) — Practical steps on configuring `recordcount`, `operationcount`, and executing `ycsb load` vs `ycsb run` against a cloud cluster.
* [[ScaleGrid MongoDB YCSB Benchmark Guide](https://scalegrid.io/blog/how-to-benchmark-mongodb-with-ycsb/)](https://scalegrid.io/blog/how-to-benchmark-mongodb-with-ycsb/) — Explains real-world configurations, including tweaking cloud parameters, adjusting thread counts, and establishing secure database authentication in performance runs.
* [[OVHcloud Public Cloud MongoDB Benchmarking](https://help.ovhcloud.com/csm/en-public-cloud-databases-mongodb-benchmark?id=kb_article_view&sysparm_article=KB0065880)](https://help.ovhcloud.com/csm/en-public-cloud-databases-mongodb-benchmark?id=kb_article_view&sysparm_article=KB0065880) — Comprehensive tutorial showing how cloud operators explicitly configure YCSB properties files to size out production tiers.

saibulusu:
Shouldn't use scenario names within the code, dropping and loading at the beginning of the profile's actions.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-YCSB.json
saibulusu: Will go through these steps a through f

brdeyo: Per our discussion:
Disregard #7. We will not target the Peak Event Shift in the profile.

We will run the workloads in the order they are noted above (A - F). Need to drop and recreate the MongoDB on each iteration of the profile so that it is a fresh database each time.

saibulusu: Iterating A through F for the workloads now.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-YCSB.json
saibulusu: Renamed.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-YCSB.json
saibulusu: Consolidated drop & load into an action at the top called initialize_database.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-YCSB.json
saibulusu: called it read50_readmodifywrite50

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-YCSB.json
saibulusu: Removed the midstream database dropping.

src/VirtualClient/VirtualClient.Main/profiles/PERF-MONGODB-YCSB.json
```diff
@@ -0,0 +1,295 @@
...
+                "FieldCount": "128",
+                "FieldLength": "128"
+            }
+        },
+        {
+            "Type": "MongoDBClientExecutor",
+            "Parameters": {
+                "Scenario": "read50_write50",
+                "Role": "Client",
+                "MetricScenario": "{Scenario}_ops{OperationCount}_{FieldCount}_{FieldLength}_th{ThreadCount}",
```
brdeyo: Use the names noted in the comments for MetricScenario and update the individual MetricCategory values to the record count.

saibulusu: Updated, using names from previous comment and metric scenario as discussed.

brdeyo (approved):

PR #731, opened 2026-06-19
Fix ASP.NET workloads crashing on Windows due to Linux-only process commands

AlexWFMS:
## Summary

`AspNetServerExecutor` and `AspNetOrchardServerExecutor` invoked Linux-only process commands unconditionally, so the ASP.NET workloads crashed on Windows even though both executors declare `[SupportedPlatforms("linux-arm64,linux-x64,win-arm64,win-x64")]`.

- `KillServerInstancesAsync` ran `pkill` and `fuser` (Unix-only). On Windows these fail to start (`Win32Exception: The system cannot find the file specified`), so the workload crash-loops until Virtual Client hits its restart-on-crash cap and the run fails with `System.AggregateException`.
- `AspNetOrchardServerExecutor.StartServerInstances` additionally launched the server via `nohup` (Unix-only) — the same failure mode.

## Fix

Make process cleanup and the Orchard server launch platform-aware:

- `KillServerInstancesAsync` — keep `pkill`/`fuser` on Unix; use `taskkill /F /IM <image>` on Windows (killing the process also frees the listening port).
- `AspNetOrchardServerExecutor.StartServerInstances` — keep `nohup` on Unix; launch the self-contained server executable directly on Windows.

The Unix code paths are unchanged. `taskkill` is consistent with the existing Windows kill path in `ProcessExtensions.SafeKill`. I intentionally did **not** switch to `ProcessManager.GetProcesses(...)` + `SafeKill`: the test `InMemoryProcessManager` does not override `GetProcesses`, so that path would call the real `Process.GetProcessesByName("dotnet")` during unit tests and could kill the test host's own `dotnet` processes.

## Testing

**Unit tests** — updated the Windows expectations in `AspNetServerExecutorTests` and `AspNetOrchardServerExecutorTests` (Linux tests unchanged). All ASP.NET tests pass (14/14).

**End-to-end on Azure VMs** — self-contained build run on a Windows and a Linux VM with `PERF-WEB-ASPNET-BOMBARDIER.json`, telemetry shipped to Kusto:

| | Windows (win-x64) | Linux (linux-x64) |
|---|---|---|
| AspNetServerExecutor | Succeeded ×21 | Succeeded ×10 |
| Bombardier result | ~76,198 req/s, ~3.43 ms avg latency (1.18M HTTP 200s/iteration) | ~72,573 req/s, ~3.53 ms |
| `pkill` / "cannot find the file" crashes | 0 | 0 |

Windows now runs the full ASP.NET Kestrel + Bombardier workload to completion. Both platforms produce near-identical benchmark results.

## Documentation

Added a **Platform Support** section to `aspnetbench.md` clarifying that the Bombardier-based profile is cross-platform (Windows + Linux) while the Wrk-based profiles are Linux-only (Wrk has no Windows build).

### DIFF
--- a/src/VirtualClient/VirtualClient.Actions.UnitTests/AspNetBench/AspNetOrchardServerExecutorTests.cs
+++ b/src/VirtualClient/VirtualClient.Actions.UnitTests/AspNetBench/AspNetOrchardServerExecutorTests.cs
@@ -86,14 +86,16 @@ namespace VirtualClient.Actions
                 await executor.ExecuteAsync(CancellationToken.None);
 
                 this.mockFixture.Tracking.AssertCommandsExecuted(true,
-                    "pkill OrchardCore",
-                    "fuser -n tcp -k 5014",
+                    "taskkill /F /IM OrchardCore\\.Cms\\.Web\\.exe",
                     $"{Regex.Escape(this.mockDotNetPackage.Path)}\\\\dotnet\\.exe publish -c Release --sc -f net9\\.0 {Regex.Escape(this.mockOrchardCorePackage.Path)}\\\\src\\\\OrchardCore\\.Cms\\.Web\\\\OrchardCore\\.Cms\\.Web\\.csproj",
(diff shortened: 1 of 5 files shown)


AlexWFMS:
Superseded by #732, which was opened from a branch on the upstream repo (the team's standard flow) instead of a personal fork. Closing this one.

PR #736, opened 2026-06-24
Fix DeathStarBench venv install to be distro-agnostic

ericavella:
## Summary
The `PERF-NETWORK-DEATHSTARBENCH` profile installed `python3.12-venv` via `LinuxPackageInstallation`. That package only exists on **Ubuntu 24.04** (which ships Python 3.12). On **Ubuntu 22.04** (Python 3.10) and other distros the package does not exist, so apt fails and the workload can't run.

## Fix
Use the generic `python3-venv` meta-package, which always resolves to the venv module for each distro's **default** Python 3 (3.8 / 3.10 / 3.12). This makes the dependency distro-agnostic and matches the package the `DeathStarBenchExecutor` already installs at runtime (`apt install python3-venv -y`).

```diff
- "Packages-Apt": "...,libssl1.0.0,python3.12-venv"
+ "Packages-Apt": "...,libssl1.0.0,python3-venv"
```

## Testing
- JSON validated.
- No remaining references to `python3.12-venv` in the source tree.


AlexWFMS (approved):
nice2

PR #749, opened 2026-07-15
Fix Redis readiness and telemetry logger null handling

ericavella:
## Summary
- support standard `ILogger` messages without an `EventContext` in the Serilog file and Event Hub telemetry loggers
- wait for every configured Redis server TCP port before publishing `ServerState`
- add focused regression coverage for logger messages and Redis readiness

## Root cause
ASP.NET API middleware emits ordinary `ILogger` messages. The file logger dereferenced a null `EventContext`, aborting otherwise-successful heartbeat responses; the Event Hub logger had the same null-context defect and emitted `EventHubTelemetryLogger.LoggingError`. Redis then remained in heartbeat polling and clients received missing state/connection failures. Redis also published state without explicitly confirming ports 6379/6380 were accepting connections.

## Validation
- focused logger tests: 6 passed
- focused Redis executor tests: 5 passed
- ARM64 VM: `qos4redis-0715-c72a` (`Standard_D2ps_v6`)
- ExperimentId: `c9fe46be-07bd-42dc-9408-b3f381b88442`
- result: 0 error events, 0 failed outcomes, 6 successful client outcomes, 296 Memtier metric rows

### DIFF
--- a/src/VirtualClient/VirtualClient.Actions.UnitTests/Redis/RedisServerExecutorTests.cs
+++ b/src/VirtualClient/VirtualClient.Actions.UnitTests/Redis/RedisServerExecutorTests.cs
@@ -100,6 +100,7 @@ namespace VirtualClient.Actions
 
                 this.fixture.Tracking.AssertCommandExecutedTimes("chmod", 1);
                 this.fixture.Tracking.AssertCommandExecutedTimes("numactl", 1);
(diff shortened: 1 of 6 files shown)


ericavella:
Closing because VM validation did not reproduce the actual QoS failure mode. The logging null reference is non-blocking under QoS Event Hub logging, and Redis connection refusals occur after successful workload metrics rather than during initial server readiness.

PR #758, opened 2026-07-21
Fix MountDisks failure on reserved/unmountable partitions

rajchaurasia03:
**Summary**

`MountDisks` fails with `"The volume at index '' does not have either a volume index nor label defined. A mount point cannot be assigned to a partition that does not have volume information associated."` on any target disk that has a reserved/metadata partition without a mountable identity. This blocks the profile before the workload (e.g. FIO) ever runs.

**Fix**

Added an `IsMountable(DiskVolume)` guard and apply it alongside the existing access-path filter in
`CreateMountPointsAsync`. A volume is considered mountable only when it has the identity its
platform's mount operation actually requires:

- **Windows** — a volume index or a drive letter.
- **Unix** — a device path (e.g. `/dev/sdc1`).

Reserved/metadata partitions have none of these and are now skipped instead of being handed to the
disk manager. Real data volumes are unaffected and still get mounted exactly once.

### DIFF
--- a/src/VirtualClient/VirtualClient.Dependencies.UnitTests/MountDisksTests.cs
+++ b/src/VirtualClient/VirtualClient.Dependencies.UnitTests/MountDisksTests.cs
@@ -266,6 +266,110 @@ namespace VirtualClient.Dependencies
             }
         }
 
+        [Test]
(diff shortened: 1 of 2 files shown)


rajchaurasia03:

imadityaa (approved):

PR #791, opened 2026-08-18
Standardize base metadata across all Virtual Client profiles

RakeshwarK:
Standardize base metadata across all Virtual Client profiles

### DIFF
--- a/src/VirtualClient/VirtualClient.Contracts/Constants.cs
+++ b/src/VirtualClient/VirtualClient.Contracts/Constants.cs
@@ -305,9 +305,7 @@ namespace VirtualClient.Contracts
         public const string SupportsIterations = nameof(SupportsIterations);
 
         /// <summary>
-        /// Metadata = RecommendedMinimumExecutionTime. Optional. Does not apply to
-        /// profiles whose runtime is determined externally (e.g. monitor profiles run
-        /// for as long as the workload profile they are paired with).
+        /// Metadata = RecommendedMinimumExecutionTime
         /// </summary>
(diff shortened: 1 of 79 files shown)


src/VirtualClient/VirtualClient.Contracts/Constants.cs
```diff
@@ -303,6 +304,32 @@ public class ProfileMetadata
...
+
+        /// <summary>
+        /// Metadata = SupportedOperatingSystems
+        /// </summary>
+        public const string SupportedOperatingSystems = nameof(SupportedOperatingSystems);
+
+        /// <summary>
+        /// The metadata properties every profile is required to define.
+        /// </summary>
+        public static readonly IReadOnlyList<string> BaseProperties = new List<string>
```
brdeyo: We don't need to force the standardization of metadata. We just want to have standard metadata in our profiles. Let's not have any coded checks at the moment.

RakeshwarK: Thanks Bryan, removed coded checks.

src/VirtualClient/VirtualClient.Contracts/Constants.cs
```diff
@@ -303,6 +303,21 @@ public class ProfileMetadata
         /// Metadata = SupportsIterations
         /// </summary>
         public const string SupportsIterations = nameof(SupportsIterations);
+
```
brdeyo: Don't think these constants are needed anymore.

nchapagain001 (approved):

brdeyo (approved):

PR #792, opened 2026-08-19
Synchronize MySQL sysbench database population across roles

ericavella:
## Summary
- Prevent the first client-side database population step from racing MySQL server configuration and remote-user creation by waiting for the server API heartbeat.
- Detect sysbench `FATAL` output during database population even when sysbench returns exit code 0.
- Update the OLTP and TPCC profiles to use `mysql-server-8.0.46-v1.zip`.
- Upgrade SSH.NET and the PR workflow actions to supported versions so repository CI can restore and run.
- The MySQL package is uploaded to `virtualclientinternal/packages`.

## Root cause
The thread-sweep profile performs database population as a client-role action, while MySQL user creation is a server-role dependency. The first 8-thread population could reach MySQL after it started listening but before `setup-database.py` created `sbtest@10.0.1.0`, producing MySQL error 1130. Later thread counts succeeded because server configuration had completed.

The observed `FATAL: error 1130` lines were from the pre-fix reproduction. They are not expected after synchronization. The explicit `FATAL` detection remains as a safety net because sysbench returned exit code 0 for that failure; without it, VirtualClient could falsely mark database population successful.

## Validation
- All 10 `SysbenchConfigurationTests` pass, including regressions that verify the server heartbeat is confirmed before remote population starts and that fatal sysbench output fails the action.
- All 21 `SshClientProxyTests` pass with SSH.NET 2026.0.0.
- Direct pre-fix VM reproduction produced repeated `FATAL` connection errors with `SYSBENCH_EXIT_CODE=0`.
- VM experiment `85c95316-8775-46ea-9895-2bc3e6a229dc` passed targeted `oltp_read_write` with exit code 0 and no fatal connection error.
- VM experiment `27093927-fa27-421f-b3ad-10a30eefc94a` passed with MySQL `8.0.46-0ubuntu0.24.04.3`: 8,303.22 transactions/sec, 166,064.44 queries/sec, 0 ignored errors, and 0 reconnects.

### DIFF
--- a/.github/workflows/pull-request-linux.yml
+++ b/.github/workflows/pull-request-linux.yml
@@ -26,10 +26,10 @@ jobs:
         # Learn more about CodeQL language support at https://aka.ms/codeql-docs/language-support
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v4
 
     - name: Setup .NET
-      uses: actions/setup-dotnet@v3
+      uses: actions/setup-dotnet@v4
       with:
         dotnet-version: 9.0.x
 
@@ -37,7 +37,7 @@ jobs:
       run: chmod -R +x *.sh
 
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v2
+      uses: github/codeql-action/init@v3
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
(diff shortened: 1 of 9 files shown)


ericavella:
Added commit 6a4ad8826 to prevent MySQL error 1130 rather than only surfacing it. Client-role database population now waits for the server API heartbeat, which is started after MySQL configuration and remote-user creation complete.

AlexWFMS (approved):

nchapagain001 (approved):

